PR #47, opened 2025-03-04
fix: log tools on LLM span in OpenAI decorator

rfriel:
`tools` should be a top-level attribute of LlmSpan, not an attribute of each Message object in the `input` of an LlmSpan.

But, the OpenAI decorator was attaching them to individual messages (in `_extract_chat_prompt`).

This causes them to not be valid galileo_core Message objects, triggering a [fallback behavior](https://github.com/rungalileo/core/blob/d5ff0dd021826ba6829d61834930113d4065681f/src/galileo_core/schemas/logging/span.py#L62) where the serialized json of the LlmSpan input field gets dumped into the content field of a single user message (see [bug ticket](https://app.shortcut.com/galileo/story/25582/openai-wrapper-uses-wrong-format-for-messages-with-tools)).

This PR:

- logs tools on LlmSpan
- logs `functions` and `function_call` on LlmSpan as well if provided, and also logs `tool_choice` in this way
  -  like `tools`, these are top-level attributes of an OpenAI request, not attributes of individual messages. See [OpenAI API docs](https://platform.openai.com/docs/api-reference/chat) for reference

`functions` is just an old name for `tools,` and `function_call` is an old name for `tool_choice`.  So:

- `functions` is logged as LlmSpan tools
- `function_call` and `tool_choice` are logged as metadata, since we don't have these on the LlmSpan model right now

---

I notice that when I log in this way, the messages look right but the tools don't show up in the UI (I was expecting them to show up in a separate box like they did in G1.0).  That's a UI issue and we can fix it separately -- the API sends the right data to UI, they just aren't displayed


ajaynayak (approved):

ajaynayak:
LGTM, thanks for identifying this and submitting the fix @rfriel

andriisoldatenko (approved):

PR #62, opened 2025-03-17
fix: LangGraph metadata errors in the Langchain handler

ajaynayak:
Also adding unit tests for the Langchain callback handler


src/galileo/utils/serialization.py
```diff
@@ -213,3 +213,28 @@ def serialize_to_str(input_data: Any) -> str:
         # Fallback if anything goes wrong
         _logger.warning(f"Serialization failed for object of type {type(input_data).__name__}")
         return ""
+
+
+def convert_to_string_dict(input_: dict) -> dict[str, str]:
```
john-weiler, line 218: unit test plz

ajaynayak: Added

andriisoldatenko (approved):

PR #74, opened 2025-03-21
fixup: move from packaging.version import Version to openai section

andriisoldatenko:


dmcwhorter (approved):

ajaynayak:
@andriisoldatenko have you tested this both with `pip install 'galileo[openai]'` and `pip install 'galileo[all]'` ? Want to make sure that adding the packaging dependency doesn't cause issues with `langchain-core`.

ajaynayak (approved):

PR #80, opened 2025-03-27
fix: Set correct openai status code for successful run

andriisoldatenko:
resolves https://app.shortcut.com/galileo/story/26598/qa2-0-logstreams-metrics-status-code-is-0-instead-of-200-for-successful-run


dmcwhorter (approved):
need to fix the failing tests before merging

PR #82, opened 2025-03-31
fix: improve error message when galileo api key is empty or invalid

andriisoldatenko:


src/galileo/utils/catch_log.py
```diff
@@ -63,7 +63,7 @@ class DecorateAllMethods:
     def __init_subclass__(cls, **kwargs: Any) -> None:
         super().__init_subclass__(**kwargs)
         for attr, f in cls.__dict__.items():
-            if attr == "__init__":
+            if attr == "__init__" or attr == "_init":
```
andriisoldatenko, line 66:
```suggestion
            if attr == "__init__":
```

dmcwhorter:
To me I think invalid credentials should be a fatal exception that halts execution from proceeding.
wdyt @ajaynayak?  do we have a point of view like "never interrupt the user code"?

andriisoldatenko:
> To me I think invalid credentials should be a fatal exception that halts execution from proceeding. wdyt @ajaynayak? do we have a point of view like "never interrupt the user code"?

yeah that's the point ☝️ , i did quick research before, and other SDKs never fails if api key is missing or wrong, we can notify user, but don't block users code IMO.

PR #83, opened 2025-03-31
fix: make No traces to flush as info

andriisoldatenko:


franz101 (approved):

src/galileo/logger.py
```diff
@@ -519,7 +519,7 @@ def flush(self) -> list[Trace]:
             List[Trace]: The list of uploaded traces.
         """
         if not self.traces:
-            self._logger.warning("No traces to flush.")
+            self._logger.info("Attempting to flush on interpreter exit...No traces to flush.")
```
ajaynayak, line 522: The "Attempting to flush on interpreter exit..." info message should be triggered in the `terminate` method (line 579)

andriisoldatenko: fixed in 15b3499

ajaynayak (approved):

PR #85, opened 2025-04-01
feat(integration): Add support for openai agents

franz101:


src/galileo/agents.py
ajaynayak, line 1: Please move this to src/galileo/handlers/openai_agents

franz101: done

franz101:
@ajaynayak @andriisoldatenko what is the wrapper /decorator to make a function not throw an exception, so I can slim the code

src/galileo/handlers/openai_agents.py
```diff
@@ -0,0 +1,333 @@
...
+from datetime import datetime
+from typing import Any, Optional
+
+from agents import TracingProcessor, tracing
+
+from galileo.decorator import galileo_context
+from galileo.logger import GalileoLogger
+from galileo.utils.serialization import convert_to_string_dict, serialize_to_str
+
+_logger = logging.getLogger("galileo.handlers.agents")
```
andriisoldatenko, line 11:
```suggestion
_logger = logging.getLogger(__name__)
```
similar as https://github.com/rungalileo/galileo-python/blob/19870ee044a2a76c0dffa5cfc1082f629b832643/src/galileo/jobs.py#L7

src/galileo/handlers/openai_agents.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+def _get_span_name(span: tracing.Span[Any]) -> str:
+    """Get name for a given OpenAI Agent span."""
+    if name := getattr(span.span_data, "name", None):
```
andriisoldatenko, line 16:
```suggestion
    if name := getattr(span.span_data, "name"):
```
`None` is already default value

src/galileo/handlers/openai_agents.py
```diff
@@ -0,0 +1,333 @@
...
+        return "AgentStep"  # More descriptive than 'task' potentially
+    elif isinstance(span.span_data, tracing.CustomSpanData):
+        return "CustomStep"
+    else:
+        # Use the type if available, otherwise fallback
+        type_name = span.span_data.type or "UnknownStep"
+        return type_name.capitalize()
+
+
+class GalileoTracingProcessor(TracingProcessor):  # pyright: ignore[reportGeneralTypeIssues]
```
andriisoldatenko, line 38:
```suggestion
class GalileoTracingProcessor(TracingProcessor):
```

andriisoldatenko: we are not using pyright, we use mypy.

src/galileo/handlers/openai_agents.py
```diff
@@ -0,0 +1,333 @@
...
+        self._flush_on_chain_end: bool = flush_on_chain_end
+
+    def on_trace_start(self, trace: tracing.Trace) -> None:
+        """Called when a trace starts. Starts a Galileo trace."""
+        try:
+            # Use trace name as input, similar to Langchain example's root node
+            self._galileo_logger.start_trace(input=trace.name or "Unnamed Agent Trace")
+            self._trace_started.add(trace.trace_id)
+            _logger.debug(f"Started Galileo trace for agent trace_id: {trace.trace_id}")
+        except Exception as e:
```
andriisoldatenko, line 58: please use `you can use galileo.utils.catch_log.warn_catch_exception decorator for this.` for this

src/galileo/handlers/openai_agents.py
```diff
@@ -0,0 +1,333 @@
...
+            _logger.error(f"Galileo: Error in on_trace_start for trace_id {trace.trace_id}: {e}", exc_info=True)
+
+    def on_trace_end(self, trace: tracing.Trace) -> None:
+        """Called when a trace ends. Concludes the Galileo trace."""
+        if trace.trace_id not in self._trace_started:
+            _logger.warning(f"Galileo: Received on_trace_end for unknown trace_id: {trace.trace_id}")
+            return
+        try:
```
andriisoldatenko, line 66: you can move it to smaller function, then you can use decorator as well instead.

src/galileo/handlers/openai_agents.py
```diff
@@ -0,0 +1,333 @@
...
+            metadata["tools"] = serialize_to_str(span.span_data.tools)
+            metadata["handoffs"] = serialize_to_str(span.span_data.handoffs)
+            metadata["output_type"] = serialize_to_str(span.span_data.output_type)
+            self._galileo_logger.add_workflow_span(
+                name=name,
+                input="Agent Step Start",  # Input often not known at start
+                metadata=convert_to_string_dict(metadata),
+                tags=tags,
+                created_at=created_at,
+            )
```
andriisoldatenko, line 103: IMO we need to make smaller methods and invoke each one here based on span_data type, it's easy to unit-test and maintain.

src/galileo/handlers/openai_agents.py
```diff
@@ -0,0 +1,333 @@
...
+        if span.trace_id not in self._trace_started:
+            _logger.warning(
+                f"Galileo: Received on_span_start for unknown trace_id: {span.trace_id}. Span: {span.span_id}"
+            )
+            # Optionally start a trace here if one isn't active, though ideally it should be.
+            # self.on_trace_start(tracing.Trace(trace_id=span.trace_id, name=f"Inferred Trace {span.trace_id}"))
+            return  # Or proceed cautiously? Let's return for now to avoid incorrect nesting.
+
+        try:
+            self._log_span_start(span)
```
andriisoldatenko, line 202: if you have decorated function, you don't need your try/except here.

tests/fixtures/openai_agents.yaml
```diff
@@ -0,0 +1,100 @@
...
+      - 929098bcfa86e523-TXL
+      Connection:
+      - keep-alive
+      Content-Type:
+      - application/json
+      Date:
+      - Mon, 31 Mar 2025 14:39:33 GMT
+      Server:
+      - cloudflare
+      Set-Cookie:
```
andriisoldatenko, line 71: for example i'm not sure we really need it here. It's more for browser.

franz101: this is the recorded requests, you can add parts of the headers to ignore. but in general this is replaying the requests

src/galileo/handlers/openai_agents.py
franz101: using DecorateAllMethods instead

PR #87, opened 2025-04-02
feat: allowing custom hosted scorers to be used

ajaynayak:
This will allow custom hosted metrics to be used in experiments

https://app.shortcut.com/galileo/story/26780/g2-0-python-and-ts-sdks-allow-users-to-use-custom-hosted-metrics-in-experiments


src/galileo/scorers.py
```diff
@@ -6,23 +6,38 @@
...
+    ScorerConfig,
     ScorerTypeFilter,
     ScorerTypeFilterOperator,
     ScorerTypes,
 )
 
 
 class Scorers(BaseClientModel):
-    def list(self):
+    def list(self, types: list[ScorerTypes] = None):
```
andriisoldatenko, line 17: I'm affraid it will fail, when i run it with Scorers().list() cause iterating over None is not allowed.

andriisoldatenko: would be nice to have a unit tests, happy to assist with mocks if needed.

ajaynayak: nice catch - fixed, will add tests

andriisoldatenko (approved):
tested change, and added unit tests.

PR #89, opened 2025-04-03
feat: Langchain async callback handler

ajaynayak:
Adding an async version of our Langchain callback handler

https://app.shortcut.com/galileo/story/27341/galileo-2-0-python-sdk-langchain-async-callback-handler


src/galileo/handlers/langchain/async_handler.py
```diff
@@ -0,0 +1,472 @@
...
+    """
+    A node in the Langchain trace.
+
+    Attributes
+    ----------
+    node_type : LANGCHAIN_NODE_TYPE
+        The type of node.
+    span_params : dict[str, Any]
+        The parameters for the span that will be created.
+    run_id : UUID
```
franz101, line 43: some general thoughts:
- thanks for the async implementation
- found the run_id quite confusing add a comment that it's similar to the node_id
- in general I would preserve the langchain type until parsing at the end.

I've noticed we pass a lot of extra params into metadata. For example we pass ls_model for the model but it should be set in the span itself.

franz101 (approved):
added some thoughts, let me know what you think.

src/galileo/handlers/langchain/async_handler.py
```diff
@@ -0,0 +1,472 @@
...
+import logging
+import time
+from typing import Any, Literal, Optional
+from uuid import UUID
+
+from galileo import galileo_context
+from galileo.logger import GalileoLogger
+from galileo.utils.serialization import EventSerializer, convert_to_string_dict, serialize_to_str
+
+_logger = logging.getLogger("galileo.handlers.langchain")
```
andriisoldatenko, line 11:
```suggestion
_logger = logging.getLogger(__name__)
```
we use it everywhere. for example https://github.com/rungalileo/galileo-python/blob/6d8224f2478855f273afb6be8e16191bb2e6bad2/src/galileo/openai.py#L98

src/galileo/handlers/langchain/async_handler.py
```diff
@@ -0,0 +1,472 @@
...
+        """
+        Log a node and its children recursively.
+
+        Parameters
+        ----------
+        node : Node
+            The node to log.
+        """
+        is_workflow_span = False
+        input = node.span_params.get("input", "")
```
andriisoldatenko, line 141:
```suggestion
        input_ = node.span_params.get("input", "")
```
cause it shadows built-in input function

src/galileo/handlers/langchain/async_handler.py
ajaynayak: Thanks Franz! My goal was to just create an async version of our existing Langchain handler to unblock one of our customers. I'm going to put most of these suggestions into a ticket that we can use to refactor both the sync and async callback handlers.

src/galileo/handlers/langchain/async_handler.py
```diff
@@ -0,0 +1,472 @@
...
+        *,
+        run_id: UUID,
+        parent_run_id: Optional[UUID] = None,
+        tags: Optional[list[str]] = None,
+        metadata: Optional[dict[str, Any]] = None,
+        **kwargs: Any,
+    ) -> Any:
+        """
+        Langchain callback when an LLM node starts.
+
```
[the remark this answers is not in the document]
ajaynayak: good idea

PR #90, opened 2025-04-03
fix: made on_tool_end recognize the tool output without the langchain metadata

ChrisTague1:
Previously Galileo's tool output would include LangChain metadata as well:

image

> Only `content` should be included, as that is what the tool returned. The rest of the fields were added by LangChain

This change makes the output correctly just `output_of_my_tool`. This is important because tool outputs are used to compute some metrics such as correctness, where having extra data could negatively impact performance.


ajaynayak:
Thanks for submitting this @ChrisTague1 ! I've incorporated your change in the following PR: https://github.com/rungalileo/galileo-python/pull/101

ajaynayak:
Released: https://github.com/rungalileo/galileo-python/releases/tag/v0.6.0

ajaynayak:
@ChrisTague1 Closing this PR, please reach out via our support channels if you run into any issues

PR #97, opened 2025-04-08
feat: error if a non existent metric is specified

andriisoldatenko:
Resolves https://app.shortcut.com/galileo/story/27772/python-sdk-no-error-if-unknown-metrics-specified-in-experiments

## how to test

Run this sample:

```
import logging
import os
from random import choice
from string import ascii_letters

from galileo.datasets import create_dataset
from galileo.experiments import run_experiment
from galileo.projects import create_project
from galileo.prompts import create_prompt_template
from galileo.resources.models import Message, MessageRole


suffix=''.join(choice(ascii_letters) for _ in range(6))
project = f"My first project {suffix}"
create_project(project)
os.environ["GALILEO_PROJECT"] = project

prompt_template = create_prompt_template(
    name=f"storyteller-prompt-{suffix}",
    project=project,
    messages=[
        Message(role=MessageRole.SYSTEM, content="You are a great storyteller."),
        Message(role=MessageRole.USER, content="Write a story about {input}")
    ]
)


dataset = create_dataset(
    name=f"story-topics-{suffix}",
    content= [{"input": "Cryptography"}, {"input": "Middle earth"}],
)

results = run_experiment(
	"travel-experiment",
	dataset=dataset,
	prompt_template=prompt_template,
	prompt_settings={
		"model_alias": "GPT-4o",
	},
    project=project,
	metrics=["Sentence Density"],
)
```

and you'll get:

```
Traceback (most recent call last):
  File "/Users/andriisoldatenko/work/rungalileo/galileo-python/downloads/run_experiment3.py", line 40, in <module>
    results = run_experiment(
  File "/Users/andriisoldatenko/work/rungalileo/galileo-python/src/galileo/experiments.py", line 256, in run_experiment
    scorer_settings = Experiments.create_run_scorer_settings(project_obj.id, experiment_obj.id, metrics)
  File "/Users/andriisoldatenko/work/rungalileo/galileo-python/src/galileo/experiments.py", line 87, in create_run_scorer_settings
    raise ValueError(f"Non-existent metric is specified `{' '.join(metrics)}`")
ValueError: Non-existent metric is specified `Sentence Density`

Process finished with exit code 1
```

### DIFF
--- a/src/galileo/experiments.py
+++ b/src/galileo/experiments.py
@@ -86,7 +86,10 @@ class Experiments(BaseClientModel):
                 unknown_metrics.append(metric)
 
         if unknown_metrics:
-            raise ValueError(f"One or more non-existent metrics are specified: `{' '.join(unknown_metrics)}`")
+            raise ValueError(
+                "One or more non-existent metrics are specified:"
+                + ", ".join(f"'{metric}'" for metric in unknown_metrics)
+            )
 
         ScorerSettings().create(project_id=project_id, run_id=experiment_id, scorers=scorers)
         return scorers
--- a/src/galileo/experiments.py
+++ b/src/galileo/experiments.py
@@ -83,6 +83,9 @@ class Experiments(BaseClientModel):
                     scorers.append(ScorerConfig.from_dict(scorer.to_dict()))
                     break
 
+        if not scorers:
+            raise ValueError(f"Non-existent metric is specified `{' '.join(metrics)}`")
+
         ScorerSettings().create(project_id=project_id, run_id=experiment_id, scorers=scorers)
         return scorers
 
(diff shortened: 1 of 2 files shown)


src/galileo/experiments.py
```diff
@@ -83,6 +83,9 @@ def create_run_scorer_settings(
                     scorers.append(ScorerConfig.from_dict(scorer.to_dict()))
                     break
 
+        if not scorers:
+            raise ValueError(f"Non-existent metric is specified `{' '.join(metrics)}`")
```
ajaynayak, line 87: Instead of outputting all of the metrics in this error message, can we just output the metrics that aren't in `all_scorers`?

andriisoldatenko: that's a good point, 
what about if user specify 1 valid and 1 invalid metric? should I fail or as implemented now - we fail only if all metrics are invalid?

ajaynayak: Should fail if at least one metric is invalid

ajaynayak (approved):

PR #99, opened 2025-04-09
fix: Use `Message` and `MessageRole` models from `galileo-core`

csurfer:
### Summary

Following PR does the following according to the detailed discussion which happened [here](https://galileoteamworkspace.slack.com/archives/C087G0X1JG2/p1744123147608439)

1. Adds wrapper classes for `Message` and `MessageRole` within python SDK which internally inherits from `galileo_core`. This is done to ensure name, functionality normalization across the board [27505](https://app.shortcut.com/galileo/story/27505/g2-0-python-sdk-standardize-the-messagerole-models)
2. Exposes the new class (along with its component classes) right at the package level so that they can be imported as `from galileo import Message, MessageRole, ToolCall, ToolCallFunction`. This is done to ensure `galileo_core` remains encapsulated and also to prevent confusion of importing from different packages/sources.
3. Adds `to_dict` method to `Message` class as the generated client makes call to `to_dict` method.

### Testing

Installed the package in edit mode via `pip install -e src/galileo-python` and ensured 
1. `MessageRole` enum works and uses lower cased members
2. `Message` role class can be instantiated and `to_dict` method is supported and works as expected. Also added a UT for the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,17 +248,17 @@ datasets = list_datasets()
 Run an experiment with a prompt template:
 
 ```python
+from galileo import Message, MessageRole
 from galileo.datasets import get_dataset
 from galileo.experiments import run_experiment
 from galileo.prompts import create_prompt_template
-from galileo.resources.models import MessageRole, Message
 
 prompt = create_prompt_template(
     name="my-prompt",
     project="new-project",
     messages=[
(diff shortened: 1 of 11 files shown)


setu4993 (approved):

ajaynayak (approved):

ajaynayak:
Nice work @csurfer !

PR #101, opened 2025-04-10
fix: Updating the Langchain handler retriever, tool, async chain callbacks

ajaynayak:
Few fixes:
1. `on_retriever_start` wasn't working correctly bc it was missing a positional argument (`serialized`).
2. `on_tool_end` was serializing unnecessary Langchain metadata as the output when we only care about the `content` (if it exists in the object). This was originally suggested by a user (@ChrisTague1) in the following PR: https://github.com/rungalileo/galileo-python/pull/90 . Since the tool output is typed as Any and Langchain doesn't have proper documentation for what to expect, I'm doing some type checking and handling.
3. In async streaming scenarios, `on_chain_start `actually sends an empty string as the input and then sends the full input in `on_chain_end`. Updating the async handler to reflect this.

### DIFF
--- a/src/galileo/handlers/langchain/async_handler.py
+++ b/src/galileo/handlers/langchain/async_handler.py
@@ -394,26 +394,38 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
         )
 
     async def on_tool_end(
-        self, output: str, *, run_id: UUID, parent_run_id: Optional[UUID] = None, **kwargs: Any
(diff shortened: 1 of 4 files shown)


setu4993 (approved):
Looks good. Would be nice to get the `langchain` callback here one of these days updated to make sure we're pulling in all of the things we've learnt for Evaluate / Observe in the past...

PR #112, opened 2025-04-29
chore: pulling the latest openapi.yaml and regenerating the API client

ajaynayak:
Ignore openapi.yaml and all of the autogenerated files in /src/galileo/resources.

I ran the e2e integration test scripts to validate this: https://github.com/rungalileo/e2e-testing/tree/main/sdk-py

Main changes:
- Fixed the generation scripts
- Updated `import-openapi-yaml.sh` to use the client API definition (not the full one). This greatly reduced the number of autogenerated client schemas and methods (as evidenced by the deletes in the changelog). This also required some changes to the client API definition. PR: https://github.com/rungalileo/api/pull/3955
- Fixed tests and client library methods post-generation (many schemas had changed)

The main issues encountered when generating the API:
1. There are duplicate `Message` and `MessageRole` schemas present in `galileo_core`. Ticket to fix that: https://app.shortcut.com/galileo/story/29814/consolidate-the-duplicate-message-and-messagerole-schemas-in-galileo-core
2. `openapi-python-client` gets angry when an endpoint has a body param schema that has a default value (ex. https://github.com/rungalileo/api/blob/7d6f18fd8039e2f6117062481150f4185419eae0/api/routers/content/dataset.py#L186 ).

We can fix the first issue via schema consolidation. I did a quick search and I wasn't able to find a solution or patch for the second one. If we're able to address that, then we can setup a Github hook/action that automatically creates PRs whenever there are client API changes (similar to the `galileo-js` repo).


andriisoldatenko (approved):
Run few basic tests: (run experiments, create datasets) and it works!

PR #116, opened 2025-04-30
feat: local scorers for local runner experiments

dmcwhorter:
Adds local scorers for locally run experiments

https://app.shortcut.com/galileo/story/24663


src/galileo/logger.py
```diff
@@ -572,6 +588,13 @@ async def async_flush(self) -> list[Trace]:
             last_output = GalileoLogger._get_last_output(current_parent)
             self.conclude(output=last_output, conclude_all=True)
 
+        if self.local_scorers:
+            self._logger.info("Computing metrics for local scorers...")
+            import asyncio
+
+            tasks = [asyncio.to_thread(populate_local_metrics, trace, self.local_scorers) for trace in self.traces]
```
dmcwhorter, line 595: @ajaynayak pls review if this is a good async pattern?

dmcwhorter:
sc-24663

src/galileo/logger.py
andriisoldatenko: @dmcwhorter we need to double check if it can work when user run python-sdk in multithreaded coded. We had issues in past. IMO we should do it seqentially or atleast tell user of SDK that we can enable asyncio.

dmcwhorter: sounds good -- do we have a test or example that I can try of where this had problems before?

andriisoldatenko:
```
@log
def _call_llm_galileo(messages: list[dict]) -> str:
    client = OpenAI()
    response = client.chat.completions.create(model="gpt-4o", messages=messages)
    return response.choices[0].message.content

messages = [{"role": "user", "content": "Hey bot"}]

# This one works fine
c = _call_llm_galileo(messages)
print("run 1", c)


# # This one does not -- even with 1 single input (works fine without the log though)
inputs = [{"messages": messages} for _ in range(2)]
with ThreadPoolExecutor(max_workers=1) as executor:
    futures = [
        executor.submit(_call_llm_galileo, **input_) for input_ in inputs
    ]

    contents = []
    for future in as_completed(futures):
        c = future.result()
        contents.append(c)
print("run 2", contents)
```

need to make sure this code works.

dmcwhorter: will verify thank you

src/galileo/decorator.py
```diff
@@ -837,7 +838,11 @@ def reset_trace_context(self) -> None:
         _trace_context.set(None)
 
     def init(
-        self, project: Optional[str] = None, log_stream: Optional[str] = None, experiment_id: Optional[str] = None
+        self,
+        project: Optional[str] = None,
+        log_stream: Optional[str] = None,
+        experiment_id: Optional[str] = None,
+        local_scorers: Optional[list[LocalScorerConfig]] = None,
```
ajaynayak, line 845: @dmcwhorter Have we decided to start using "scorers" instead of "metrics"? For ex. run_experiment has a `metrics` input param, but we're using `local_scorers` here

ajaynayak: In order to maintain consistency, we could change this input param to `metrics` and only accept LocalScorerConfig (or LocalMetricConfig) initially. I assume that we might want to support adding hosted metrics to traces at some point in the future?

src/galileo/decorator.py
```diff
@@ -225,6 +227,7 @@ def log(
         name: Optional[str] = None,
         span_type: Optional[SPAN_TYPE] = None,
         params: Optional[dict[str, Union[str, Callable]]] = None,
+        dataset_record: Optional[DatasetRecord] = None,
```
dmcwhorter, line 230: @ajaynayak does this seem like a reasonable pattern for now to propagate the dataset record?

ajaynayak: LGTM, just add a docstring calling out its usage in experiments

src/galileo/logger.py
```diff
@@ -104,11 +107,16 @@ class GalileoLogger(TracesLogger, DecorateAllMethods):
...
 
     _logger = logging.getLogger("galileo.logger")
 
     def __init__(
-        self, project: Optional[str] = None, log_stream: Optional[str] = None, experiment_id: Optional[str] = None
+        self,
+        project: Optional[str] = None,
+        log_stream: Optional[str] = None,
+        experiment_id: Optional[str] = None,
+        local_scorers: Optional[list[LocalScorerConfig]] = None,
```
ajaynayak, line 119: `local_metrics` ?

src/galileo/utils/singleton.py
```diff
@@ -76,7 +77,12 @@ def _get_key(
         return key + (project, log_stream)
 
     def get(
-        self, *, project: Optional[str] = None, log_stream: Optional[str] = None, experiment_id: Optional[str] = None
+        self,
+        *,
+        project: Optional[str] = None,
+        log_stream: Optional[str] = None,
+        experiment_id: Optional[str] = None,
+        local_scorers: Optional[list[LocalScorerConfig]] = None,
```
ajaynayak, line 85: `local_metrics` ?

src/galileo/decorator.py
```diff
@@ -485,7 +491,7 @@ def _get_span_param_names(self, span_type: SPAN_TYPE) -> list[str]:
         Returns:
             List of parameter names that can be used with the specified span type
         """
-        common_params = ["name", "input", "metadata", "tags"]
+        common_params = ["name", "input", "user_metadata", "tags"]
```
ajaynayak, line 494: We surface this field as `metadata` in logger.py

pyproject.toml
```diff
@@ -11,13 +11,14 @@ dynamic = ["dependencies"]
...
 wrapt = "^1.14"
 attrs = ">=22.2.0"
 python-dateutil = "^2.8.0"
 langchain-core = { version = "^0.3.14", optional = true }
-openai = { version = "^1.66.5", optional = true }
-openai-agents = { version = "^0.0.12", optional = true }
+openai = { version = "<1.67.0", optional = true }
+openai-agents = { version = "<0.0.13", optional = true }
+
+galileo-core = "~=3.32.0"
```
john-weiler, line 21: @ajaynayak -- did we resolve the issue with >3.29.0 ?

src/galileo/datasets.py
```diff
@@ -486,24 +487,33 @@ def get_dataset_version(
...
+        if not isinstance(output, str):
+            output = json.dumps(output)
+
+    metadata = {}
+    if "metadata" in values_dict and values_dict["metadata"]:
+        if isinstance(values_dict["metadata"], str):
+            try:
+                metadata = json.loads(values_dict["metadata"])
+            except json.decoder.JSONDecodeError:
+                raise ValueError("Metadata field must be a serialized json object with string values")
```
john-weiler, line 510: Can you say where this metadata field is derived from? IE: datasets or some other more specific info

src/galileo/datasets.py
```diff
@@ -486,24 +487,33 @@ def get_dataset_version(
...
         else:
-            result.append(row_values)
-
-    return result
+            metadata = values_dict["metadata"]
+        if not isinstance(metadata, dict):
+            raise ValueError("Metadata must be a dictionary")
```
john-weiler, line 514: same here re the error

src/galileo/experiments.py
```diff
@@ -158,17 +164,20 @@ def run_with_function(
...
+        galileo_context.init(project=project_obj.name, experiment_id=experiment_obj.id, local_metrics=local_metrics)  # type: ignore[arg-type]
 
-        logged_process_func = log(name=experiment_obj.name)(func)
+        def logged_process_func(row: DatasetRecord) -> Callable:
+            return log(name=experiment_obj.name, span_type=span_type, dataset_record=row)(func)
 
         #  process each row in the dataset
         for row in records:
-            results.append(process_row(row, logged_process_func))
+            results.append(process_row(row, logged_process_func(row)))
```
john-weiler, line 180: nice.

src/galileo/experiments.py
```diff
@@ -309,48 +322,47 @@ def _load_dataset_and_records(
...
+    if id:
+        dataset = get_dataset(id=id)
+        if not dataset:
+            raise ValueError("Could not find dataset with id " + id)
+    elif name:
+        dataset = get_dataset(name=name)
+        if not dataset:
+            raise ValueError("Could not find dataset with name " + name)
+    else:
+        raise ValueError("One of id or name must be provided")
```
john-weiler, line 352: This one also could be more informative.

src/galileo/schema/metrics.py
```diff
@@ -0,0 +1,32 @@
...
+class LocalMetricConfig(BaseModel, Generic[MetricType]):
+    name: str = Field(description="Name of the local metric")
+    scorer_fn: Callable[[Union[Trace, Span]], MetricType] = Field(
+        description="function to call to produce the metric value (takes a trace or span as input"
+    )
+    aggregator_fn: Optional[Callable[[list[MetricType]], Union[MetricType, dict[str, MetricType]]]] = Field(
+        default=None,
+        description="function to call to produce the aggregate metric values from individual metric values",
+    )
+    scorable_types: list[StepType] = Field(default=[StepType.llm])
```
john-weiler, line 23: this should be scoreable_node_types ?

john-weiler: Also instead of StepType, this should be NodeType ?

src/galileo/schema/metrics.py
```diff
@@ -0,0 +1,32 @@
...
+    aggregatable_types: list[StepType] = Field(default=[StepType.trace, StepType.workflow])
```
john-weiler, line 24: do we support aggregatable_types anywhere else ?

src/galileo/schema/metrics.py
dmcwhorter: No, we don't use the concept of "nodes" at all in G2.0 traces/spans.  Traces & Spans (& sessions) inherit from BaseStep and contain a `type` field that uses `StepType`

src/galileo/datasets.py
dmcwhorter: this whole function is dataset values, this is the dataset metadata field, one of 3 supported columns in datasets:
1. input
2. output
3. metadata

src/galileo/schema/metrics.py
dmcwhorter: I don't think so.  How does this work with code-based scorers?  Is it just the root/trace that gets aggregations?
Eventually we'll need to account for sessions.

src/galileo/decorator.py
```diff
@@ -843,8 +862,13 @@ def init(
             project: The project name. Defaults to None.
             log_stream: The log stream name. Defaults to None.
             experiment_id: The experiment id. Defaults to None.
+            local_metrics: Local metrics configs to run on the traces/spans before submitting them for ingestion.  Defaults to None.
         """
         GalileoLoggerSingleton().reset(project=project, log_stream=log_stream, experiment_id=experiment_id)
+        GalileoLoggerSingleton().get(
```
andriisoldatenko, line 868: @dmcwhorter i'm not sure that I understand why we do this line here, to me seems redundant 🤔

dmcwhorter: because when `get` is actually called, we don't have `local_metrics` to set the metrics for the context.

src/galileo/utils/singleton.py
```diff
@@ -76,7 +77,12 @@ def _get_key(
         return key + (project, log_stream)
 
     def get(
-        self, *, project: Optional[str] = None, log_stream: Optional[str] = None, experiment_id: Optional[str] = None
+        self,
+        *,
+        project: Optional[str] = None,
+        log_stream: Optional[str] = None,
+        experiment_id: Optional[str] = None,
+        local_metrics: Optional[list[LocalMetricConfig]] = None,
```
andriisoldatenko, line 85: so we want to cache logger instance using `local_metrics` as well?

dmcwhorter: yes correct, so they get computed on `flush`

dmcwhorter: oh, no, we don't want the local_metrics in the cache key if that's what you're asking.  They should be set at `init` but not used to access the logger

src/galileo/decorator.py
andriisoldatenko: true,

andriisoldatenko (approved):

ajaynayak (approved):

PR #126, opened 2025-05-07
bugfix: Using context var stacks for saving context with nested _call_ s

pratduv:


src/galileo/logger.py
```diff
@@ -516,9 +516,8 @@ def conclude(
 
         # TODO: Allow the final span output to propagate to the parent spans
         current_parent = None
-        if conclude_all:
-            while self.current_parent() is not None:
-                current_parent = super().conclude(output=output, duration_ns=duration_ns, status_code=status_code)
+        while self.current_parent() is not None:
```
ajaynayak, line 519: Why was the conclude_all check removed?

ajaynayak:
@pratduv Nice work. Can you add one or more tests that cover the nested behavior that you're trying to fix?

src/galileo/logger.py
pratduv: Because right above we check if not conclude_all -> do XYZ, so by line 519 conclude_all = true

tests/test_nested_context.py
ajaynayak, line 1: Please add these tests to `test_decorator.py`

ajaynayak (approved):

PR #146, opened 2025-05-27
chore: updating the api client to add a Protect route

ajaynayak:
1) I updated my local `api` client_api.py to pull in the Protect route and then 2) re-generated openapi.yaml using `import-openapi-yaml.sh` and then 3) re-generated the api client methods using `auto-generate-api-client.sh`.


src/galileo/resources/api/data/list_all_versions_for_scorer_scorers_scorer_id_versions_get.py
```diff
@@ -0,0 +1,191 @@
...
+        starting_token=starting_token,
+        limit=limit,
+    )
+
+    response = await client.get_async_httpx_client().request(**kwargs)
+
+    return _build_response(client=client, response=response)
+
+
+async def asyncio(
```
andriisoldatenko, line 162: is it accurate? why the name is `asyncio` 🤔 ?

PR #152, opened 2025-05-29
chore: Test codeflash integration on new code

csurfer:
Codeflash had some issues running pytest as we distribute the tests across CPU cores using xdist. This PR 
* added some intentionally bad code
* validated that we are able to optimize it via codeflash with all the necessary changes and then 
* removed the bad code leaving the PR with just the setup changes.


pyproject.toml
```diff
@@ -144,6 +144,7 @@ omit = [
 module-root = "./src/galileo"
 tests-root = "tests"
 test-framework = "pytest"
+pytest-cmd = "pytest -n 0"
```
misrasaurabh1, line 147:
```suggestion
```

csurfer: Done.

ajaynayak (approved):

PR #170, opened 2025-06-12
feat: Pass in metric version and name to experiments

john-weiler:


john-weiler:
https://app.shortcut.com/galileo/story/32279/sdk-support-for-setting-metric-versions

src/galileo/schema/metrics.py
```diff
@@ -33,3 +34,23 @@ def set_aggregatable_types(cls, value: list[StepType], info: ValidationInfo) ->
             if step_type not in [StepType.workflow, StepType.trace]:
                 raise ValidationError("aggregatable_types can only contain trace or workflow steps")
         return value
+
+
+class Metric(BaseModel):
+    name: Union[str] = Field(
+        description="The name of the metric you want to run a specific version of (ie: 'Sentence Density')."
+    )
+    version: Union[int, None] = Field(
```
dmcwhorter, line 43: maybe `Optional[int]` would be clearer

src/galileo/experiments.py
```diff
@@ -88,27 +88,43 @@ def list(self, project_id: str) -> Optional[Union[HTTPValidationError, list["Exp
...
+
+        all_scorers = Scorers().list()
+        known_metrics = {metric.name: metric for metric in all_scorers}
+
+        unknown_metrics = []
+
+        for metric in metrics:
+            if isinstance(metric, LocalMetricConfig):
+                local_metric_configs.append(metric)
+                continue
```
dmcwhorter, line 105: this continue is unnecessary right?

src/galileo/experiments.py
```diff
@@ -88,27 +88,43 @@ def list(self, project_id: str) -> Optional[Union[HTTPValidationError, list["Exp
...
+                version = metric.version if isinstance(metric, Metric) else None
+
+                if name in known_metrics:
+                    raw_metric_dict = known_metrics[name].to_dict()
+
+                    # Set the version on the ScorerConfig if provided
+                    if version is not None:
+                        raw_version = Scorers().get_scorer_version(scorer_id=raw_metric_dict["id"], version=version)
+                        raw_metric_dict["scorer_version"] = raw_version.to_dict()
+                    scorers.append(ScorerConfig.from_dict(raw_metric_dict))
```
dmcwhorter, line 117: this seems a bit weird to me that we have to retrieve the whole scorer version info and then set it in the config before creating the experiment.  I would have expected that we could just create the experiment with the right version and retrieving the scorer config for the version etc would happen on the backend.

john-weiler: Yeah I agree - however , the problem is all the scorer setting stuff happens in Promptgalileo, which doesn’t have access to API, and so without completely refactoring that part I just had us pass in the whole version for now. Can talk about refactoring later I think

dmcwhorter: ok yeah makes sense

PR #177, opened 2025-06-16
feat: refactor GalileoLogger to split GalileoBatch abd GalileoStream accordingly

andriisoldatenko:
This PR refactor GalileoLogger and introduce new parameter `mode`.

Also it splits main logic between 2 classes `GalileoBatchLogger` and `GalileoStreamingLogger` which both implements `GalileoInterface`.

resolves


src/galileo/logger/iface.py
```diff
@@ -0,0 +1,57 @@
+import abc
+from datetime import datetime
+from typing import Optional
+
+from galileo_core.schemas.logging.span import StepWithChildSpans
+from galileo_core.schemas.logging.step import StepAllowedInputType
+from galileo_core.schemas.logging.trace import Trace
+
+
+class GalileoInterface(metaclass=abc.ABCMeta):
```
ajaynayak, line 10: Can we rename this to `IGalileoLogger` and this file to interface.py?

tests/test_langchain.py
```diff
@@ -10,16 +10,16 @@
 from galileo import Message, MessageRole
 from galileo.handlers.langchain import GalileoCallback
 from galileo.handlers.langchain.handler import _root_node
-from galileo.logger import GalileoLogger
+from galileo.logger.logger import GalileoLogger
```
ajaynayak, line 13: `GalileoLogger` should be importable via `from galileo import GalileoLogger` or `from galileo.logger import GalileoLogger`

src/galileo/logger/batch.py
```diff
@@ -0,0 +1,133 @@
...
+from galileo.utils.nop_logger import nop_sync
+from galileo_core.schemas.logging.span import StepWithChildSpans
+from galileo_core.schemas.logging.step import StepAllowedInputType
+from galileo_core.schemas.logging.trace import Trace
+from galileo_core.schemas.shared.traces_logger import TracesLogger
+
+
+class GalileoBatchLogger(TracesLogger, GalileoInterface):
+    """
+    Galileo Batch logger
```
ajaynayak, line 18: Please add a more descriptive docstring. Also mention that you shouldn't use this directly (use `GalileoLogger` with `mode` instead)

src/galileo/logger/stream.py
```diff
@@ -0,0 +1,5 @@
+from galileo.logger.iface import GalileoInterface
+
+
+class GalileoStreamLogger(GalileoInterface):
```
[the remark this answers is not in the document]
andriisoldatenko: fixed

ajaynayak (approved):
Nice work!

PR #179, opened 2025-06-16
feat: making Session name optional; adding set_session to galileo_context

ajaynayak:


src/galileo/schema/trace.py
```diff
@@ -23,14 +23,14 @@ class TracesIngestResponse(BaseLogStreamOrExperimentModel):
...
-    name: str = Field(..., description="Name of the session.")
+    name: Optional[str] = Field(default=None, description="Name of the session.")
     previous_session_id: Optional[UUID4] = Field(default=None, description="Previous session id.")
     external_id: Optional[str] = Field(default=None, description="External id of the session.")
 
 
 class SessionCreateResponse(BaseLogStreamOrExperimentModel):
     id: UUID4 = Field(description="Id of the session.")
-    name: str = Field(description="Name of the session.")
+    name: Optional[str] = Field(default=None, description="Name of the session.")
```
pratduv, line 33: When are we creating the session name (if not provided by the user?) Is it created with the log stream insights - and so will only be created when user requests insights?

andriisoldatenko (approved):

PR #182, opened 2025-06-17
feat(core): Add Protect to the python SDK

jrhone:
**Shortcut:** https://app.shortcut.com/galileo/story/30923/add-the-protect-tool-to-the-2-0-python-sdk

**Description:** Adds `Protect` invoke and `Stage` get/create/update/pause/resume.

**Tests:**
- [ ] Local.
- [ ] CI.


src/galileo/__init__.py
```diff
@@ -5,7 +5,10 @@
 
 from galileo.decorator import GalileoDecorator, galileo_context, log
 from galileo.logger import GalileoLogger
+from galileo.resources.models.execution_status import ExecutionStatus
```
setu4993, line 8: Why are we using auto-gen schemas instead of the one from `galileo-core`?

setu4993 (changes requested):

src/galileo/__init__.py
jrhone: I pushed the changes for Protect calls to use the galileo-core schemas.

The Stage calls require a bit of extra work as I have to move some schemas from api to core
- https://github.com/rungalileo/core/pull/322
- https://github.com/rungalileo/api/pull/4329
- https://github.com/rungalileo/galileo-python/pull/185

src/galileo/constants/protect.py
```diff
@@ -0,0 +1,3 @@
+from datetime import timedelta
+
+TIMEOUT = timedelta(seconds=10).total_seconds()
```
ajaynayak, line 3: change to TIMEOUT_SECS and set to a integer value instead of computing on import

ajaynayak (approved):
Nice work!

src/galileo/constants/protect.py
jrhone: Done! (on my [other branch](https://github.com/rungalileo/galileo-python/pull/196/files))

PR #184, opened 2025-06-18
feat: add Python client functions for global prompt template endpoints

shriyaedu:
**Shortcut:** 
https://app.shortcut.com/galileo/story/31564/add-new-python-client-functions

**Description:**  
This PR adds new SDK utilities for managing **global** prompt templates. All functions operate at the global level and do not require a project name.

---

### `get_prompts(name, limit=100)`

Fetches a list of **global** prompt templates filtered by name.

- You **must** provide a `name` string to filter by (exact match).
- Returns up to `limit` templates (default: 100).

**Parameters:**

| Name    | Type     | Description                                       |
|---------|----------|---------------------------------------------------|
| `name`  | `string` | Filter by template name (required)                |
| `limit` | `number` | Max number of templates to return (default: 100)  |

**Returns:** `list[PromptTemplate]`

---

### `get_prompt({ id?, name?, version? })`

Retrieves a specific **global** prompt template version.

- You **must** provide **either** `id` or `name`, but **not both**.
- If `version` is omitted, the currently selected version is returned.
- If `version` is provided with `name`, the ID is resolved automatically.

**Parameters:**

| Name      | Type     | Description                                                      |
|-----------|----------|------------------------------------------------------------------|
| `id`      | `string` | Prompt template ID (optional, required if `name` is not provided)|
| `name`    | `string` | Prompt template name (optional, required if `id` is not provided)|
| `version` | `number` | Version number to retrieve (optional)                            |

**Returns:** `PromptTemplateVersion | None`  
**Raises:** `ValueError` if neither or both `id` and `name` are provided.

---

### `create_prompt(name, template)`

Creates a new **global** prompt template.

**Parameters:**

| Name       | Type                      | Description                                                  |
|------------|---------------------------|--------------------------------------------------------------|
| `name`     | `string`                  | Name of the new template                                     |
| `template` | `string | list[Message]`  | Template content as plain string or list of `Message` objects|

**Returns:** `PromptTemplate`  
**Raises:** `PromptTemplateAPIException` on API failure.

---

### `delete_prompt({ id?, name? })`

Deletes a **global** prompt template.

- You must provide **either** `id` or `name`, but **not both**.
- If `name` is provided, the template ID will be resolved automatically.

**Parameters:**

| Name   | Type     | Description                                      |
|--------|----------|--------------------------------------------------|
| `id`   | `string` | Prompt template ID (optional if `name` is given) |
| `name` | `string` | Prompt template name (optional if `id` is given) |

**Returns:** `None`  
**Raises:** `ValueError` if neither or both are provided, or if not found.

---

### `get_prompt_templates(project_name, name)`

> **Deprecated** – Use `get_prompts(name)` instead.

Fetches prompt templates **scoped to a project**.

**Parameters:**

| Name           | Type     | Description                |
|----------------|----------|----------------------------|
| `project_name` | `string` | Project to scope the search|
| `name` | `string` | Prompt template name|

**Returns:** `list[PromptTemplate]`

---

### `get_prompt_template({ id?, name?, project_name, version? })`

> **Deprecated** – Use `get_prompt({ id?, name?, version? })` instead.

Fetches a specific **project-scoped** prompt template version.

**Parameters:**

| Name           | Type     | Description                        |
|----------------|----------|------------------------------------|
| `id`           | `string?`| Prompt template ID                 |
| `name`         | `string?`| Prompt template name               |
| `project_name` | `string` | Scope to a project                 |
| `version`      | `number?`| Version number to retrieve         |

**Returns:** `PromptTemplate`

---

### `create_prompt_template(name, template, project_name)`

> **Deprecated** – Use `create_prompt(name, template)` instead.

Creates a **project-scoped** prompt template.

**Parameters:**

| Name           | Type                      | Description                                                 |
|----------------|---------------------------|-------------------------------------------------------------|
| `name`         | `string`                  | Name of the new template                                    |
| `template`     | `string \| list[Message]` | Prompt content (string or list of `Message` objects)        |
| `project_name` | `string`                  | Project scope                                               |

**Returns:** `PromptTemplate`

**Tests:**
- [x] Local.
- [ ] CI.


src/galileo/prompts.py
```diff
@@ -128,3 +262,39 @@ def get_prompt_template(project: str, name: str) -> Optional[PromptTemplate]:
...
+
+@overload
+def delete_template(*, name: str) -> None: ...
+
+
+def delete_template(*, template_id: Optional[str] = None, name: Optional[str] = None) -> None:
+    return GlobalPromptTemplates().delete(template_id=template_id, name=name)  # type: ignore[call-overload]
+
+
+def create_template(name: str, template: Union[list[Message], str]) -> Optional[PromptTemplate]:
```
ajaynayak, line 291: Users will be confused between `create_template` and `create_prompt_template`. Can we rename the global methods to `create_global_prompt_template` , etc. ? Please add docstrings as well

jonahe, line 291: Why is the return type optional?

shriyaedu: Good catch, creating a prompt template should always return a PromptTemplate object. I changed it!

src/galileo/prompts.py
```diff
@@ -128,3 +262,144 @@ def get_prompt_template(project: str, name: str) -> Optional[PromptTemplate]:
...
+    Returns
+    -------
+    Optional[PromptTemplateVersion]
+        The template version if found, None otherwise.
+    """
+    return GlobalPromptTemplates().get_version(template_id=template_id, version=version)
+
+
+def list_global_prompt_templates(
+    name_filter: Optional[str] = None, limit: Union[Unset, int] = 100
```
vamaq, line 377: I think you are missing the `starting_token` parameter here, right?

shriyaedu: I didn't include the starting token in the global method to keep it consistent with the datasets' methods. My reasoning for this was that users might typically choose 0 as the starting token since this isn't a UI based interaction, and they may not need pagination when they are using the SDK. But if you think we should provide users with the option to specify a starting token, I can add this to the global endpoint.

src/galileo/prompts.py
```diff
@@ -127,4 +264,159 @@ def get_prompt_template(project: str, name: str) -> Optional[PromptTemplate]:
...
+@overload
+def get_prompt(*, template_id: str) -> Optional[PromptTemplateVersion]: ...
+
+
+@overload
+def get_prompt(*, name: str) -> Optional[PromptTemplateVersion]: ...
+
+
+def get_prompt(
+    *, template_id: Optional[str] = None, name: Optional[str] = None, version: Optional[int] = None
```
vamaq, line 284: Considering that we are removing the word template from the method we should probably change the name of this parameter to just `id` and also update all the related documentation.

src/galileo/prompts.py
```diff
@@ -127,4 +264,159 @@ def get_prompt_template(project: str, name: str) -> Optional[PromptTemplate]:
...
+            if not parent_template:
+                return None
+            target_id = parent_template.id
+
+        if not target_id:
+            raise ValueError("A template id is required to fetch a specific version.")
+
+        return GlobalPromptTemplates().get_version(template_id=target_id, version=version)
+
+    prompt_template: Optional[PromptTemplate] = None
```
vamaq, line 327: Considering that you are already validating that either the id or the name should be present you can simplify this code to just:

```suggestion
    if template_id:
        prompt_template = GlobalPromptTemplates().get(template_id=template_id)
    else:
        prompt_template = GlobalPromptTemplates().get(name=name)
```

PR #196, opened 2025-06-24
feat(protect): integrated changes

jrhone:
**Shortcut**: https://app.shortcut.com/galileo/story/30923/add-the-protect-tool-to-the-2-0-python-sdk

Protect Tool
- Adds a custom Langchain tool for Protect calls that allows Protect data to appear under Traces in the UI
- Adds support for `stage_version` as an input to the tool which wasn't in the Protect 1.0 SDK

Schema Translation
- Uses core schemas in the client layer and translates to generated schemas in the api layer
- Manually translates `rulesets` to `prioritized_rulesets` as the generated schema expects it

### DIFF
--- a/src/galileo/protect.py
+++ b/src/galileo/protect.py
@@ -30,22 +30,6 @@ class Protect(BaseClientModel, DecorateAllMethods):
         metadata: Optional[dict[str, str]] = None,
         headers: Optional[dict[str, str]] = None,
     ) -> Optional[Union[Response, HTTPValidationError]]:
-        """
-        Calls invoke api.
-
-        Returns
-        -------
-        Response
-            Various data from api
-
-        Raises
-        ------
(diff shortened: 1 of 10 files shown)


src/galileo/protect.py
```diff
@@ -56,10 +58,16 @@ def invoke(
             metadata=metadata,
             headers=headers,
         )
+        request_dict = request.model_dump()
+        request_dict["prioritized_rulesets"] = request_dict.pop("rulesets", [])
```
jrhone, line 62: Note: translation required where `RulesetsMixin` is used since generated model uses `prioritized_rulesets`

ajaynayak (approved):

jrhone:
This work will be merged in via https://github.com/rungalileo/galileo-python/pull/198

PR #198, opened 2025-06-24
feat(core): Add Protect to the python SDK

jrhone:
**Shortcut**: https://app.shortcut.com/galileo/story/30923/add-the-protect-tool-to-the-2-0-python-sdk

Protect SDK
- Adds Protect `invoke` and `ainvoke` to the client
- Adds Stage `get/create/update/pause/resume` to the client

Protect Tool
- Adds a custom Langchain tool for Protect calls that allows Protect data to appear under Traces in the UI.
- Adds support for `stage_version` as an input to the tool which wasn't in the Protect 1.0 SDK

Schema Translation
- Uses core schemas in the client layer and translates to generated schemas in the api layer
- Manually translates `rulesets` to `prioritized_rulesets` as the generated schema expects it.

Documentation
- Updated docstrings to be Google style and only expose the helper methods

This is a combination of the following PRs, pointed against main with the newly generated types
https://github.com/rungalileo/galileo-python/pull/182
https://github.com/rungalileo/galileo-python/pull/196


src/galileo/handlers/langchain/tool.py
```diff
@@ -0,0 +1,83 @@
...
+class ProtectTool(BaseTool):
+    name: str = "GalileoProtect"
+    description: str = (
+        "Protect your LLM applications from harmful content using Galileo Protect. "
+        "This tool is a wrapper around Galileo's Protect API, can be used to scan text "
+        "for harmful content, and can be used to trigger actions based on the results."
+        "The tool can be used synchronously or asynchronously, on the input text or output text,"
+        "and can be configured with a set of rulesets to evaluate on."
+    )
+    args_schema: type[BaseModel] = ProtectToolInputSchema
```
setu4993, line 30: IIRC this needs to be `pydantic` v1 and can't inherit from `pydantic.BaseModel`.

src/galileo/protect.py
```diff
@@ -0,0 +1,106 @@
...
+from galileo.resources.models.request import Request as APIRequest
+from galileo.resources.models.response import Response as APIResponse
+from galileo.utils.catch_log import DecorateAllMethods
+from galileo_core.schemas.protect.payload import Payload
+from galileo_core.schemas.protect.request import Request
+from galileo_core.schemas.protect.response import Response
+from galileo_core.schemas.protect.ruleset import Ruleset
+
+
+class Protect(BaseClientModel, DecorateAllMethods):
```
setu4993, line 19: Why are we wrapping this in a class?

src/galileo/protect.py
```diff
@@ -0,0 +1,106 @@
...
+        prioritized_rulesets=prioritized_rulesets,
+        project_id=project_id,
+        project_name=project_name,
+        stage_id=stage_id,
+        stage_name=stage_name,
+        stage_version=stage_version,
+        timeout=timeout,
+        metadata=metadata,
+        headers=headers,
+    )
```
setu4993, line 106: No async?

setu4993 (commented):
This mostly looks good. I asked a couple questions.

src/galileo/handlers/langchain/tool.py
jrhone: Oh I see.. yea it's `pydantic` v1 in the Protect SDK. Could you share why? I'll add a comment to the line.

src/galileo/protect.py
jrhone: Mainly following the pattern we use in the other modules, but the class also lets us pull in the functionality from `BaseClientModel` and `DecorateAllMethods`

src/galileo/protect.py
jrhone: It doesn't seem like we have async methods in the client for any of the other calls so hesitated adding one here.

setu4993: We'll need async for Protect. Most of our customers use the async variants.

jrhone: Cool beans, I added `async invoke` to Protect and `async run` to the Protect Tool following the pattern in the `galileo-protect` repo.

PR #205, opened 2025-06-27
fix: Accounting for agent type in langchain handler

pratduv:
Shortcut: https://app.shortcut.com/galileo/story/34184/langchain-smith-integration-identify-agent-span

### DIFF
--- a/src/galileo/handlers/langchain/async_handler.py
+++ b/src/galileo/handlers/langchain/async_handler.py
@@ -104,7 +104,7 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
         node : Node
             The node to log.
         """
-        is_workflow_span = False
+        is_span_with_children = False
         input_ = node.span_params.get("input", "")
         output = node.span_params.get("output", "")
         name = node.span_params.get("name")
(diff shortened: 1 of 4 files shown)


src/galileo/handlers/langchain/async_handler.py
```diff
@@ -190,8 +202,8 @@ async def _log_node_tree(self, node: Node) -> None:
             else:
                 _logger.warning(f"Child node {child_id} not found")
 
-        # Conclude workflow span. Use the last child's output if necessary
-        if is_workflow_span:
+        # Conclude parent span. Use the last child's output if necessary
+        if is_span_with_children:
```
pratduv, line 206: @ajaynayak minor comment here but was wondering if I could use `SPAN_TYPES_WITH_CHILD_SPANS` , but then we have to make sure that node.node_type will always line up with `StepType`.

ajaynayak: Not necessary right now IMO

franz101 (commented):
Looks good, this requires UI changes too?

dmallon (approved):

PR #212, opened 2025-07-03
feat(core): Add option project_id and log_stream_id to GalileoLogger instantiation

dmallon:


src/galileo/logger/logger.py
```diff
@@ -124,7 +123,9 @@ class GalileoLogger(GalileoBatchLogger, GalileoStreamingLogger, DecorateAllMetho
     def __init__(
         self,
         project: Optional[str] = None,
+        project_id: Optional[str] = None,
```
setu4993, line 126: These should be UUIDs, no?

src/galileo/logger/logger.py
```diff
@@ -134,35 +135,52 @@ def __init__(
         project_name_from_env = getenv("GALILEO_PROJECT", DEFAULT_PROJECT_NAME)
         log_stream_name_from_env = getenv("GALILEO_LOG_STREAM", DEFAULT_LOG_STREAM_NAME)
 
+        project_id_from_env = getenv("GALILEO_PROJECT_ID")
+        log_stream_id_from_env = getenv("GALILEO_LOG_STREAM_ID")
+
         self.project_name = project or project_name_from_env
+        self.project_id = project_id or project_id_from_env
```
setu4993, line 142: Unrelated, but this is exactly where using `galileo_core.BaseConfig` would simplify a ton of this logic. We allow init from env vars (type-cast with Pydantic) or as args, and it all just works...

src/galileo/logger/logger.py
```diff
@@ -134,35 +135,52 @@ def __init__(
...
+        if self.project_name is None and self.project_id is None:
             raise GalileoLoggerException(
-                "User must provide project_name to GalileoLogger, or set it as an environment variable."
+                "User must provide project_name or project_id to GalileoLogger, or set it as an environment variable."
             )
 
-        if log_stream and experiment_id:
-            raise GalileoLoggerException("User must provide either experiment_id or log_stream, not both.")
+        if (log_stream or log_stream_id) and experiment_id:
+            raise GalileoLoggerException("User cannot specify both a log stream and an experiment.")
```
setu4993, line 150: This validation should occur after L156, no? Until then, we don't really _know_ if we have been given both.

setu4993 (commented):
Mostly looks good. A couple questions.

src/galileo/logger/logger.py
ajaynayak: The GalileoCoreApiClient has project_id and log_stream_id as str, so maintaining that

setu4993 (approved):
This looks good for now.

I do think there's a bunch of clean up we could do if we leverage how we use `GalileoConfig` from `galileo-core` and how we simplify a ton of this logic in our other Python libraries.

PR #218, opened 2025-07-07
fix: Raise error on Job Create failure for Run Experiments

jrhone:
**Shortcut:** https://app.shortcut.com/galileo/story/33156/if-experiment-creation-fails-with-the-python-sdk-the-console-shows-that-it-worked

**Description:** The `create job` during `run experiment` was failing with an error message due to a misconfigured model integration, yet the experiment printed a 'started' message afterwards. This raises the error instead of printing it.

I think `create job` is only used by Experiments so this change shouldn't affect anything else. We raise a ValueError similar to the other errors raised by `run experiment` caused by a user issue.

**Tests:**

- [ ] Unit Tests Added


src/galileo/jobs.py
```diff
@@ -33,5 +33,5 @@ def create(
         _logger.info(f"create job: {create_params}")
         result = create_job_jobs_post.sync_detailed(client=self.client, body=CreateJobRequest(**create_params))
         if not result.parsed or not isinstance(result.parsed, CreateJobResponse):
-            _logger.error(f"create job failed: {result.content}")
+            raise ValueError(f"create job failed: {result.content}")
```
setu4993, line 36: I don't think we should raise a `ValueError` here... It should be an `GalileoHTTPException`.

ajaynayak (approved):

PR #226, opened 2025-07-12
feat: Adding streaming support to the GalileoLogger

ajaynayak:
Streaming mode allows the client to send events to our API as they happen, as opposed to explicitly flushing in batch at the end of a trace. The usage is the same, except for the addition of a `mode` attribute in the GalileoLogger constructor:

```
logger = GalileoLogger(mode="streaming")
logger.start_trace(input="Streaming Logger Test")
logger.add_agent_span(input="test", agent_type="planner")
logger.add_llm_span(
    input={"role": "user", "content": "test"},
    output="test",
    model="gpt-4o",
    num_input_tokens=10,
    num_output_tokens=3,
    total_tokens=13,
    duration_ns=1000,
    tools=tools,
)
logger.conclude(output="agent end", duration_ns=2000) # conclude the agent span
logger.conclude(output="test", duration_ns=2000) # conclude the trace

# no flush required!
```

Streaming mode utilizes our `EventLoopThreadPool` from galileo-core, which allows multiple API requests to be handled simultaneously with retries.

When the Python interpreter closes, instead of flushing like in batch mode, we wait to ensure all running thread pool tasks are completed.


src/galileo/logger/logger.py
```diff
@@ -161,6 +176,10 @@ def __init__(
         if local_metrics:
             self.local_metrics = local_metrics
 
+        if self.mode == "streaming":
+            self._max_retries = STREAMING_MAX_RETRIES
```
pratduv, line 180: Do we want users to be able to set this via an env variable? and we cap it to some x number

src/galileo/logger/logger.py
```diff
@@ -223,6 +242,152 @@ def _get_last_output(node: Union[BaseStep, None]) -> Optional[str]:
...
 
+    @nop_sync
+    def _ingest_trace_streaming(self, trace: Trace, is_complete: bool = False) -> None:
+        traces_ingest_request = TracesIngestRequest(
+            traces=[copy.deepcopy(trace)], session_id=self.session_id, is_complete=is_complete, reliable=True
+        )
+
+        task_id = f"trace-ingest-{trace.id}"
+
+        @backoff.on_exception(
```
pratduv, line 253: Nice!!

src/galileo/logger/logger.py
```diff
@@ -339,8 +511,16 @@ def add_single_llm_span_trace(
             dataset_output=dataset_output,
             dataset_metadata=dataset_metadata,
             span_step_number=span_step_number,
+            trace_id=uuid.uuid4(),
+            span_id=uuid.uuid4(),
         )
 
+        if self.mode == "streaming":
+            self.traces = [trace]
+            self._ingest_step_streaming(trace, is_complete=True)
```
pratduv, line 520: So in streaming mode, if someone were to call `add_singe_llm_span_trace`, we want to mark is_complete as True? Why did we go this route?

pratduv: Hm I guess this makes sense if we want to not leave the "trace" as incomplete even during streaming mode, and if users wanted to add multiple spans to an existing trace they should've used `start_trace` and `add_xxx` methods. Can we leave a comment here detailing this?

pratduv: Now I wonder what the rationale  was to add the `add_single_llm_span_to_trace` method in the first place to the SDK

src/galileo/logger/logger.py
```diff
@@ -601,7 +810,31 @@ def add_agent_span(
...
+        current_parent = self.current_parent()
+        if current_parent is None:
+            raise ValueError("No existing workflow to conclude.")
+
+        current_parent.output = output or current_parent.output
+        current_parent.status_code = status_code
+        if duration_ns is not None:
+            current_parent.metrics.duration_ns = duration_ns
+
+        finished_step = self._parent_stack.pop()
```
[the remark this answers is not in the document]
pratduv: If the ValueError is raised it'll be popped but not returned.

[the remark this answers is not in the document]
pratduv: Looks like we also have this issue in `traces_logger.py`

src/galileo/logger/logger.py
```diff
@@ -656,18 +931,61 @@ async def async_flush(self) -> list[Trace]:
...
+            self.flush()
+        else:
+            print("Checking if all tasks are completed...")
+            while not self._task_handler.all_tasks_completed():
+                time.sleep(0.1)
+            self._task_handler.terminate()
+            print("All tasks completed. Exiting...")
+
+        end_time = time.perf_counter()
+        print(f"time taken in terminate: {(end_time - start_time):0.4f} seconds")
```
pratduv, line 988: Is it convention to print this out?

pratduv:
Looking good!

Why did we initially go with having two separate loggers, and why did we choose to store everything in one logger but have two modes instead?

src/galileo/logger/logger.py
ajaynayak: `add_single_llm_span_trace` is a convenience method that's part of the `galileo-core` TracesLogger that we inherit from, so we're supporting it. We can remove it in the future if it's not really used.

src/galileo/logger/logger.py
ajaynayak: Oops, I need to remove this

ajaynayak:
> Looking good!
> 
> Why did we initially go with having two separate loggers, and why did we choose to store everything in one logger but have two modes instead?

We weren't able to get a clean abstraction of having two inherited loggers with shared code and differences in the way that they handle object operations including conclude(). The implementation was cleaner with a single logger and a bit of branching logic on `mode`. Check out Andrii's earlier commits for a full history.

tests/test_logger_batch.py
```diff
@@ -108,7 +108,16 @@ def test_single_span_trace_to_galileo(
             )
         ],
     )
-    assert payload == expected_payload
+
+    trace = payload.traces[0]
+    assert trace.input == expected_payload.traces[0].input
```
dmcwhorter, line 113: why did we have to go to doing this one field at a time?

dmcwhorter (approved):
lgtm.  My main feedback is i wish that we didn't have to have a custom async task pool implementation as a part of this

src/galileo/schema/trace.py
```diff
@@ -69,3 +143,8 @@ class LogRecordsSearchFilter(BaseModel):
 
 class LogRecordsSearchRequest(BaseLogStreamOrExperimentModel):
     filters: Optional[list[LogRecordsSearchFilter]] = Field(default=None, description="Filters to apply to the search.")
+
+
+RetrieverSpanAllowedOutputType = Union[
```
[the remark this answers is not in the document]
ajaynayak: This doesn't exist in core - it expects a Document[] in the output field. The Python logger has a more relaxed schema for the output field (to make it a bit more user-friendly) and has a helper method to convert it to a list of Documents. This type declaration was in logger.py before and was moved to schema/trace.py. It should probably be moved to `core`.

src/galileo/logger/utils.py
```diff
@@ -19,3 +24,35 @@ def get_last_output(node: Union[BaseStep, None]) -> Optional[str]:
...
+                print("Trace not found, retrying...")
+                raise e
+            if e.status_code == 408:
+                print("Request timed out, retrying...")
+                raise e
+            if e.status_code == 429:
+                print("Rate limited, retrying...")
+                raise e
+            if e.status_code >= 500:
+                print("Server error, retrying...")
```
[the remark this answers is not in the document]
ajaynayak: Should be > 500, nice catch

src/galileo/logger/logger.py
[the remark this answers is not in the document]
ajaynayak: The finished step needs to be popped from the stack bc the operation has been concluded. if a ValueError is raised, then nothing is returned because the logger state is malformed.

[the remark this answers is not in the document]
pratduv: Maybe they forgot to `conclude()` one last span in a complex heirarchy. I can imagine in a notebook environment for example folks might just want to try again. Since we're not doing too many things before we flush the span (we're just popping from the parent stack AFAIR), maybe we can restore state?

ajaynayak:
> Approved! Please take a look at these comments before merging https://github.com/rungalileo/galileo-python/pull/226/files#r2223172304 https://github.com/rungalileo/galileo-python/pull/226/files/1b109960e20d3f4e5ba8575e12293cabb2cce77f#r2228504009

Took a look, thanks!

PR #249, opened 2025-07-23
fix: Make langchain callbacks produce data closer to our schema's intentions

rfriel:
I was looking over some recent logstreams created with Langchain/Langgraph today, and noticed that a lot of them contained "raw Langchain" data in span inputs and/or outputs.

E.g. LLM messages that have `"content"` as a dict with its own inner "content" key, plus a dict called "additional_kwargs," and all of that stuff.  And we often weren't setting our own schema's `tool_call_id` either on tool nodes or on the LLM output messages that called the tools.

----

This PR updates our langchain callbacks to fix at least some of the problems I observed.

I tested this against the [langgraph-fsi-agent](https://github.com/rungalileo/sdk-examples/tree/main/python/agent/langgraph-fsi-agent) example from our SDK examples repo, and I can confirm it works in that case, although I can't guarantee it works everywhere.  It does change the SDK's behavior, so users might be confused at least for a little while even if it's ultimately a good change.

I'll document what changed in comments.  Note that we have two callback classes, the sync one and the async one, which are largely identical; I tried to keep both in sync, let me know if you see any differences.

---

EDIT: also made a change that fetches more informative names for chain nodes, see [slack discussion](https://galileoteamworkspace.slack.com/archives/C087G0X1JG2/p1753317258496139?thread_ts=1753316118.299229&cid=C087G0X1JG2)

### DIFF
--- a/src/galileo/handlers/langchain/async_handler.py
+++ b/src/galileo/handlers/langchain/async_handler.py
@@ -309,7 +309,7 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
             return
 
         node_type = "chain"
-        node_name = GalileoCallback._get_node_name(node_type, serialized)
+        node_name = GalileoCallback._get_node_name(node_type, serialized, kwargs)
 
         # If the `name` is `LangGraph` or `Agent`, set the node type to `agent`.
         if node_name in ["LangGraph", "Agent"]:
@@ -350,7 +350,7 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
         Note: This callback is only used for non-chat models.
         """
         node_type = "llm"
-        node_name = GalileoCallback._get_node_name(node_type, serialized)
+        node_name = GalileoCallback._get_node_name(node_type, serialized, kwargs)
         invocation_params = kwargs.get("invocation_params", {})
         model = invocation_params.get("model_name", "")
         temperature = invocation_params.get("temperature", 0.0)
@@ -393,7 +393,7 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
     ) -> Any:
         """Langchain callback when a chat model starts."""
         node_type = "chat"
-        node_name = GalileoCallback._get_node_name(node_type, serialized)
+        node_name = GalileoCallback._get_node_name(node_type, serialized, kwargs)
         invocation_params = kwargs.get("invocation_params", {})
         model = invocation_params.get("model", invocation_params.get("_type", "undefined-type"))
         temperature = invocation_params.get("temperature", 0.0)
@@ -455,7 +455,7 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
     ) -> Any:
         """Langchain callback when a tool node starts."""
         node_type = "tool"
-        node_name = GalileoCallback._get_node_name(node_type, serialized)
+        node_name = GalileoCallback._get_node_name(node_type, serialized, kwargs)
         if "inputs" in kwargs and isinstance(kwargs["inputs"], dict):
             input_str = json.dumps(kwargs["inputs"], cls=EventSerializer)
         await self._start_node(
@@ -468,43 +468,12 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
             metadata={k: str(v) for k, v in metadata.items()} if metadata else None,
         )
 
-    @staticmethod
-    def _find_tool_message(obj: Any) -> Optional[ToolMessage]:
-        """
-        Look for a ToolMessage in the output passed to langchain's on_tool_end callback.
-        If found, return the ToolMessage. Otherwise, return None.
-
-        Can return a ToolMessage in two cases:
-        1. The output is already a ToolMessage.
-        2. The output is a Command object with a ToolMessage in its update list.
-
-        As of this writing, Command and ToolMessage are the only Langchain/Langgraph
-        classes inheriting from ToolOutputMixin. And Langchain is supposed to convert
-        any tool output **not** inheriting from ToolOutputMixin to a ToolMessage.
-
-        So this should cover all cases. Either:
-
-        - the output is a Command (converted to ToolMessage here),
-        - or Langchain will force it to be a ToolMessage before passing it to us.
-        """
-        if isinstance(obj, ToolMessage):
-            return obj
-        if hasattr(obj, "update") and isinstance(obj.update, dict) and "messages" in obj.update:
-            update_messages = obj.update["messages"]
-            if (
-                isinstance(update_messages, list)
-                and len(update_messages) > 0
-                and isinstance(update_messages[-1], ToolMessage)
-            ):
-                return update_messages[-1]
-        return None
-
     async def on_tool_end(
         self, output: Any, *, run_id: UUID, parent_run_id: Optional[UUID] = None, **kwargs: Any
     ) -> Any:
         """Langchain callback when a tool node ends."""
         end_node_kwargs = {}
-        if (tool_message := self._find_tool_message(output)) is not None:
+        if (tool_message := GalileoCallback._find_tool_message(output)) is not None:
             end_node_kwargs["output"] = tool_message.content
             end_node_kwargs["tool_call_id"] = tool_message.tool_call_id
         else:
@@ -531,7 +500,7 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
     ) -> Any:
         """Langchain callback when a retriever node starts."""
         node_type = "retriever"
-        node_name = GalileoCallback._get_node_name(node_type, serialized)
+        node_name = GalileoCallback._get_node_name(node_type, serialized, kwargs)
         await self._start_node(
             node_type,
             parent_run_id,
--- a/src/galileo/handlers/langchain/async_handler.py
+++ b/src/galileo/handlers/langchain/async_handler.py
@@ -17,7 +17,7 @@ try:
     from langchain_core.agents import AgentAction, AgentFinish
     from langchain_core.callbacks.base import AsyncCallbackHandler
     from langchain_core.documents import Document
-    from langchain_core.messages import BaseMessage
+    from langchain_core.messages import BaseMessage, ToolMessage
     from langchain_core.outputs import LLMResult
 except ImportError:
     _logger.warning("Failed to import langchain, using stubs")
@@ -27,6 +27,7 @@ except ImportError:
     BaseMessage = object
     LLMResult = object
     AgentFinish = object
+    ToolMessage = object
 
 
 class GalileoAsyncCallback(AsyncCallbackHandler):
@@ -188,6 +189,7 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
                 tags=tags,
                 created_at=created_at,
                 step_number=step_number,
+                tool_call_id=node.span_params.get("tool_call_id"),
             )
         else:
             _logger.warning(f"Unknown node type: {node.node_type}")
@@ -468,34 +470,45 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
         **kwargs: Any,
     ) -> Any:
         """Langchain callback when a tool node starts."""
-        print(f"on_tool_start serialized: {type(serialized)}, {serialized}")
-        print(f"on_tool_start input_str: {type(input_str)}, {input_str}")
         node_type = "tool"
         node_name = GalileoCallback._get_node_name(node_type, serialized)
+        if "inputs" in kwargs and isinstance(kwargs["inputs"], dict):
+            input_str = json.dumps(kwargs["inputs"], cls=EventSerializer)
         await self._start_node(
             node_type,
             parent_run_id,
             run_id,
             name=node_name,
-            input=json.dumps(json.loads(input_str), cls=EventSerializer),
+            input=input_str,
             tags=tags,
             metadata={k: str(v) for k, v in metadata.items()} if metadata else None,
         )
+
+    @staticmethod
+    def _find_tool_message(obj: Any) -> Optional[ToolMessage]:
+        if isinstance(obj, ToolMessage):
+            return obj
+        if hasattr(obj, "update") and isinstance(obj.update, dict) and "messages" in obj.update:
+            update_messages = obj.update["messages"]
+            if isinstance(update_messages, list) and len(update_messages) > 0 and isinstance(update_messages[-1], ToolMessage):
+                return update_messages[-1]
+        return None
 
     async def on_tool_end(
         self, output: Any, *, run_id: UUID, parent_run_id: Optional[UUID] = None, **kwargs: Any
     ) -> Any:
         """Langchain callback when a tool node ends."""
         print(f"on_tool_end output: {type(output)}, {output}")
-        output_data = json.loads(json.dumps(output, cls=EventSerializer))
-        print(f"on_tool_end output_data: {type(output_data)}, {output_data}")
-        # if isinstance(output, dict) and "content" in output:
-        #     output = serialize_to_str(output["content"])
-        # elif hasattr(output, "content"):
-        #     output = serialize_to_str(output.content)
-        # else:
-        #     output = serialize_to_str(output)
-        await self._end_node(run_id, output=output_data["content"], tool_call_id=output_data["tool_call_id"])
+        end_node_kwargs = {}
+        if (tool_message := self._find_tool_message(output)) is not None:
+            print(f"on_tool_end tool_message: {tool_message}")
+            end_node_kwargs["tool_call_id"] = tool_message.tool_call_id
+            end_node_kwargs["output"] = json.dumps(tool_message.content, cls=EventSerializer)
+        else:
+            print(f"on_tool_end no tool message found")
+            end_node_kwargs["output"] = json.dumps(output, cls=EventSerializer)
+        print(f"on_tool_end end_node_kwargs: {end_node_kwargs}")
+        await self._end_node(run_id, **end_node_kwargs)
 
     async def on_retriever_start(
         self,
--- a/src/galileo/handlers/langchain/async_handler.py
+++ b/src/galileo/handlers/langchain/async_handler.py
@@ -474,7 +474,11 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
             return obj
         if hasattr(obj, "update") and isinstance(obj.update, dict) and "messages" in obj.update:
             update_messages = obj.update["messages"]
-            if isinstance(update_messages, list) and len(update_messages) > 0 and isinstance(update_messages[-1], ToolMessage):
+            if (
+                isinstance(update_messages, list)
+                and len(update_messages) > 0
+                and isinstance(update_messages[-1], ToolMessage)
+            ):
                 return update_messages[-1]
         return None
 
--- a/src/galileo/handlers/langchain/async_handler.py
+++ b/src/galileo/handlers/langchain/async_handler.py
@@ -473,15 +473,15 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
         """
         Look for a ToolMessage in the output passed to langchain's on_tool_end callback.
         If found, return the ToolMessage. Otherwise, return None.
-        
+
         Can return a ToolMessage in two cases:
         1. The output is already a ToolMessage.
         2. The output is a Command object with a ToolMessage in its update list.
-        
+
         As of this writing, Command and ToolMessage are the only Langchain/Langgraph
         classes inheriting from ToolOutputMixin. And Langchain is supposed to convert
         any tool output **not** inheriting from ToolOutputMixin to a ToolMessage.
-        
+
         So this should cover all cases. Either:
 
         - the output is a Command (converted to ToolMessage here),
--- a/src/galileo/handlers/langchain/async_handler.py
+++ b/src/galileo/handlers/langchain/async_handler.py
@@ -391,6 +391,8 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
         **kwargs: Any,
     ) -> Any:
         """Langchain callback when a chat model starts."""
+        # print(f"on_chat_model_start serialized: {type(serialized)}, {serialized}")
+        # print(f"on_chat_model_start messages: {type(messages)}, {messages}")
         node_type = "chat"
         node_name = GalileoCallback._get_node_name(node_type, serialized)
         invocation_params = kwargs.get("invocation_params", {})
@@ -400,11 +402,18 @@ class GalileoAsyncCallback(AsyncCallbackHandler):
 
         # Serialize messages safely
         try:
-            flattened_messages = [message.dict() for batch in messages for message in batch]
-            serialized_messages = json.loads(json.dumps(flattened_messages, cls=EventSerializer))
+            for batch in messages:
+                for message in batch:
+                    print(type(message))
+            # flattened_messages = [message.model_dump() for batch in messages for message in batch]
+            flattened_messages = [message for batch in messages for message in batch]
+            prepped_messages = flattened_messages
+            _logger.info(f"Messages before serialization: {prepped_messages}")
+            serialized_messages = json.loads(json.dumps(prepped_messages, cls=EventSerializer))
         except Exception as e:
             _logger.warning(f"Failed to serialize chat messages: {e}")
             serialized_messages = str(messages)
+        _logger.info(f"Messages after serialization: {serialized_messages}")
 
         await self._start_node(
             node_type,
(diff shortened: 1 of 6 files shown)


src/galileo/handlers/langchain/async_handler.py
```diff
@@ -14,19 +14,19 @@
 _logger = logging.getLogger(__name__)
 
 try:
-    from langchain_core.agents import AgentAction, AgentFinish
```
rfriel, line 17: `AgentAction` was unused

src/galileo/handlers/langchain/async_handler.py
```diff
@@ -188,6 +188,7 @@ async def _log_node_tree(self, node: Node) -> None:
                 tags=tags,
                 created_at=created_at,
                 step_number=step_number,
+                tool_call_id=node.span_params.get("tool_call_id"),
```
rfriel, line 191: We weren't including `tool_call_id` in the ToolSpan creation kwargs, even if was added to span_params correctly in one of the callback methods.

src/galileo/handlers/langchain/async_handler.py
```diff
@@ -400,7 +401,7 @@ async def on_chat_model_start(
 
         # Serialize messages safely
         try:
-            flattened_messages = [message.dict() for batch in messages for message in batch]
```
rfriel, line 403: #220  replaced `.dict()` with `.model_dump()` in the sync version of the handler, but the async one was still using `.dict()`.

Here, I don't do either one, I just pass the raw messages into `EventSerializer,` which already knows how to handle them.

src/galileo/handlers/langchain/async_handler.py
```diff
@@ -455,6 +456,8 @@ async def on_tool_start(
         """Langchain callback when a tool node starts."""
         node_type = "tool"
         node_name = GalileoCallback._get_node_name(node_type, serialized)
+        if "inputs" in kwargs and isinstance(kwargs["inputs"], dict):
```
rfriel, line 459: The `input_str` that langchain passes to `on_tool_start` is difficult to interpret, since it often comes from calling `str()` on some python object (see [here](https://github.com/langchain-ai/langchain/blob/a7d0e42f3fa5b147fea9109f60e799229f30a68b/libs/core/langchain_core/tools/base.py#L826)) and so you can't always parse it as json.  In principle, `ast.literal_eval` could do it if you had all the right imports in scope already, but I don't want to go there.

However, I noticed that langchain can also pass `inputs`, which is [the same thing as input_str  _before_ the str() call if the input is a dict (and None otherwise)](https://github.com/langchain-ai/langchain/blob/a7d0e42f3fa5b147fea9109f60e799229f30a68b/libs/core/langchain_core/tools/base.py#L834).

Here, I use `inputs` if provided and hand it along to EventSerializer, which will try to serialize its contents in the "usual" way (i.e. the way we would serialize them if they had appeared elsewhere in the callback).

src/galileo/handlers/langchain/async_handler.py
```diff
@@ -465,17 +468,48 @@ async def on_tool_start(
             metadata={k: str(v) for k, v in metadata.items()} if metadata else None,
         )
 
+    @staticmethod
+    def _find_tool_message(obj: Any) -> Optional[ToolMessage]:
```
rfriel, line 472: This is a workaround for langchain-supervisor's [create_handoff_tool](https://github.com/langchain-ai/langgraph-supervisor-py/blob/74a4eca7f3051ccba448e7593eb32db09bf77919/langgraph_supervisor/handoff.py#L55-L125), which returns a `Command` object.

It seems that `Command` is the one thing that langchain will allow returning as-is from a tool without trying to convert it into a ToolMesssage.  Here, I extract "the actual ToolMessage" from the Command object if we get one.

src/galileo/utils/serialization.py
```diff
@@ -43,6 +43,11 @@ def _serialize_zoned_datetime(v: dt.datetime) -> str:
         return _serialize_zoned_datetime(localized_dt)
 
 
+def map_langchain_role(role: str) -> str:
```
rfriel, line 46: Langchain's eccentric "ai" and "human" roles aren't in [our MessageRole enum](https://github.com/rungalileo/core/blob/15ce57617c40e6dabfabed9d09e09ce7f6fc06fe/src/galileo_core/schemas/logging/llm.py#L7-L14), so they block otherwise-acceptable messages from getting interpreted as galileo-core Message objects.  Here I map them to the galileo MessageRole equivalents.

src/galileo/utils/serialization.py
```diff
@@ -81,20 +83,12 @@ def default(self, obj: Any) -> Any:
             if isinstance(obj, date):
                 return obj.isoformat()
 
-            elif isinstance(obj, BaseModel):
```
rfriel, line 84: I moved a bunch of stuff up or down within this function to make the conditional logic match the class hierarchy. 

That is, so that we reach the block that handles a subclass before we reach a block that handles its base class. (Otherwise we'd never hit the block for the subclass, even when the subclass is passed in.)

So, we check for more specific things like AIMessage first, and only later do we check for stuff like Dataclass and BaseModel.

src/galileo/utils/serialization.py
```diff
@@ -81,20 +83,12 @@ def default(self, obj: Any) -> Any:
...
-                return self.default(
-                    obj.model_dump(mode="json", exclude_none=True, exclude_unset=True, exclude_defaults=True)
-                )
             if isinstance(obj, Path):
                 return str(obj)
 
             if is_langchain_available:
-                from langchain_core.load.serializable import Serializable
-
-                if isinstance(obj, Serializable):
```
rfriel, line 94: Some (all?) of the langchain objects tested in this section were `Serializable`s, so the code blocks for them were never being reached.

src/galileo/utils/serialization.py
```diff
@@ -105,25 +99,36 @@ def default(self, obj: Any) -> Any:
...
-                    # Map the `type` to `role`.
-                    dumped = obj.model_dump(include={"content", "type"})
-                    dumped["role"] = dumped.pop("type")
-                    return self.default(dumped)
-                elif isinstance(obj, (AIMessageChunk, AIMessage)) and obj.tool_calls:
+                elif isinstance(obj, (AIMessageChunk, AIMessage)):
                     # Map the `type` to `role`.
-                    dumped = obj.model_dump(include={"content", "type", "tool_calls"})
-                    dumped["role"] = dumped.pop("type")
+                    dumped = obj.model_dump(include={"content", "type", "additional_kwargs"})
```
rfriel, line 104: Langchain puts tool_calls in "additional_kwargs", so we have to look for it there.

src/galileo/utils/serialization.py
```diff
@@ -105,25 +99,36 @@ def default(self, obj: Any) -> Any:
...
                     dumped = obj.model_dump(include={"content", "type", "status", "tool_call_id"})
-                    dumped["role"] = dumped.pop("type")
+                    dumped["role"] = map_langchain_role(dumped.pop("type"))
+                    return self.default(dumped)
+                elif isinstance(obj, BaseMessage):
+                    # Map the `type` to `role`.
+                    dumped = obj.model_dump(include={"content", "type"})
+                    dumped["role"] = map_langchain_role(dumped.pop("type"))
                     return self.default(dumped)
-                elif isinstance(obj, BaseModel):
```
rfriel, line 123: This was duplicative of the BaseModel check above (now removed), and also Serializable inherits from BaseModel so we have to check Serializable before checking BaseModel if we want both cases to get used.

src/galileo/utils/serialization.py
```diff
@@ -81,49 +83,60 @@ def default(self, obj: Any) -> Any:
...
-
-                if isinstance(obj, Serializable):
-                    return obj.to_json()
-
-                from langchain_core.agents import AgentAction, AgentFinish
                 from langchain_core.messages import AIMessage, AIMessageChunk, BaseMessage, ToolMessage
                 from langchain_core.outputs import ChatGeneration, LLMResult
                 from langchain_core.prompt_values import ChatPromptValue
 
-                if isinstance(obj, (AgentFinish, AgentAction, ChatPromptValue)):
```
rfriel, line 102: We have tests for `AgentFinish` that expect the callback to pass through its attributes as-is rather than [using the `.messages` property](https://github.com/langchain-ai/langchain/blob/a7d0e42f3fa5b147fea9109f60e799229f30a68b/libs/core/langchain_core/agents.py#L167-L170).

The only reason this line didn't cause a problem with `AgentFinish` before now is that it was never reached in the `AgentFinish` case because it's a subclass of `Serializable`, and we had the Serializable check before this part.

src/galileo/handlers/langchain/async_handler.py
```diff
@@ -505,10 +505,17 @@ async def on_tool_end(
...
         end_node_kwargs = {}
         if (tool_message := self._find_tool_message(output)) is not None:
+            end_node_kwargs["output"] = tool_message.content
             end_node_kwargs["tool_call_id"] = tool_message.tool_call_id
-            end_node_kwargs["output"] = json.dumps(tool_message.content, cls=EventSerializer)
         else:
-            end_node_kwargs["output"] = json.dumps(output, cls=EventSerializer)
+            end_node_kwargs["output"] = output
+        # This will pass-through strings like 'blah' without encoding them
+        # into '"blah"', but will turn everything else into a JSON string.
```
rfriel, line 513: According to langchain, [`BaseMessage.content` has type `Union[str, list[Union[str, dict]]]
`](https://github.com/langchain-ai/langchain/blob/a7d0e42f3fa5b147fea9109f60e799229f30a68b/libs/core/langchain_core/messages/base.py#L26).  So either it's just a string, or it's some nested thing we have to json dump.

tests/test_langchain.py
```diff
@@ -163,7 +164,7 @@ def test_on_chain_start_with_kwargs_serialised_none(self, callback: GalileoCallb
         assert traces[0].spans[0].type == "agent"
         assert (
             traces[0].spans[0].input
-            == '{"messages": [{"content": "What does Lilian Weng say about the types of agent memory?"}]}'
+            == '{"messages": [{"content": "What does Lilian Weng say about the types of agent memory?", "role": "user"}]}'
```
rfriel, line 167: We now get "role" here, which is a good thing

tests/test_langchain.py
```diff
@@ -399,7 +398,9 @@ def __init__(self, content, tool_call_id, status, role):
         assert str(run_id) in callback._nodes
         assert callback._nodes[str(run_id)].node_type == "tool"
         assert callback._nodes[str(run_id)].span_params["input"] == "2+2"
-        assert callback._nodes[str(run_id)].span_params["output"] == "tool response"
+        assert callback._nodes[str(run_id)].span_params["output"] == (
```
rfriel, line 401: I don't think we should be grabbing "content" from some arbitrary non-langchain object (and ignoring the rest) just because it has that attribute defined...

This is not how you're supposed to log tool outputs in langchain, but if you _do_ want to pass in such a thing, we should probably pass it through and assume you want whatever you put in rather than guessing your intent. 

(Details: langchain expects a ToolMessage (or a Command) for tool outputs, and if you pass a custom object like this it will attach the object as the `content` of a ToolMessage, which we will now grab and use apart from the rest of the ToolMessage. But we aren't descending further into `tool_message.content.content`, nor should we do so)

src/galileo/handlers/langchain/async_handler.py
pratduv: >ast.literal_eval could do it if you had all the right imports in scope already, but I don't want to go there.
I'm just curious - wdym by right imports? Is it not just `import ast`?

src/galileo/utils/serialization.py
pratduv: for ref https://python.langchain.com/docs/concepts/messages/

pratduv: Do we also want to account for these types here in the EventSerializer? I don't see [HumanMessage](https://github.com/rungalileo/galileo-python/blob/7d7ead5f4b74dd609ba2373bcb481e494f1bdf85/src/galileo/utils/serialization.py#L102-L125).

src/galileo/handlers/langchain/async_handler.py
```diff
@@ -469,13 +472,20 @@ async def on_tool_end(
         self, output: Any, *, run_id: UUID, parent_run_id: Optional[UUID] = None, **kwargs: Any
     ) -> Any:
         """Langchain callback when a tool node ends."""
-        if isinstance(output, dict) and "content" in output:
-            output = serialize_to_str(output["content"])
-        elif hasattr(output, "content"):
-            output = serialize_to_str(output.content)
+        end_node_kwargs = {}
+        if (tool_message := GalileoCallback._find_tool_message(output)) is not None:
```
pratduv, line 476: Nice!

pratduv (changes requested):
Few qs/comments! Nice!!

src/galileo/handlers/langchain/handler.py
```diff
@@ -308,8 +312,13 @@ def _get_node_name(node_type: LANGCHAIN_NODE_TYPE, serialized: Optional[dict[str
                 return node_name
             elif node_class_reference and isinstance(node_class_reference, list):
                 return node_class_reference[-1]
-            else:
-                return node_type.capitalize()
+            elif isinstance(kwargs, dict):
+                if kwargs_name := kwargs.get("name"):
+                    return kwargs_name
+                if "metadata" in kwargs and isinstance(kwargs["metadata"], dict):
```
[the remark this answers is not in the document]
pratduv: Please introduce this nesting in kwargs["metadata"] in our tests as well.

src/galileo/handlers/langchain/async_handler.py
rfriel: Oh I was mistaken, ast.literal_eval can only interpret python literals like str, int, dict, so it wouldn't work in this case anyway.

I had thought you could use it to parse stuff with `HumanMessage` or other user-defined classes, but only if those classes were defined in the current scope (by importing them).  Sort of like how pickle works with imports, although pickle does it automatically.  But no, that was all something I "hallucinated" anyway

src/galileo/utils/serialization.py
pratduv: Interesting! Are we currently covering examples across all such blocks in our tests?

src/galileo/utils/serialization.py
rfriel: Your link is to the main branch. On this PR branch [I call map_langchain_role inside EventSerializer](https://github.com/rungalileo/galileo-python/blob/a1a3b9fead4535c1d12029bf6ece186da3bce69e/src/galileo/utils/serialization.py#L97-L120)

In the code I just linked, HumanMessage and SystemMessage get handled by the `elif isinstance(obj, BaseMessage):` block.

We don't need to distinguish them from their base class because they don't need any special unique handling distinct to their type (unlike, say, ToolMessage where we need to keep tool_call_id).  All of these objects have a `type` attribute which redundantly expresses which type of message they are, and we just convert SystemMessage's type of "system" to the system role, etc.

src/galileo/utils/serialization.py
rfriel: No, and I don't think we should add them as part of this PR.

It's not like we have a spec written out for the serializer that's like "if the type is list[dict[str, Foo[x[int]]]] then it should do blah, whereas if it's dict[str, Foo[Bar[str]]] we should do some other thing, etc."  I guess we _could_ write a spec like that but it would just re-express the same complex logic in a different format.

Note that langchain itself doesn't have any spec like that for how their own objects and callback-callers must behave, which is part of why our serializer and callback here are so complex -- langchain is bizarre, changing and unpredictable and so we have to catch a million edge cases

A much more valuable kind of test (IMO) would be more of an "integration test" where we actually import langchain and/or langgraph, run a "realistic" chain or agent end-to-end (mocking out the LLM API calls), and then verify that the output matches some reasonable expectation. That would guard against breaking changes as we upgrade langchain, and would be really cool, but seems conceptually distinct from this PR

src/galileo/utils/serialization.py
rfriel: Added some test code [here](https://github.com/rungalileo/galileo-python/pull/249/files/a1a3b9fead4535c1d12029bf6ece186da3bce69e..843117be61ccad9ae7d7d92a64e1582f72e84305) that checks for the system, user and assistant roles from SystemMessage, HumanMessage and AIMessage respectively

ToolMessage's role is already being checked by some of the tool tests

pratduv (approved):
Approved! Please see if it's reasonable to 1) add tests to check for "name" in kwargs["metadata"] and 2) run a few nbs (maybe @sid2harthgupta can also run his langchain agent) and we can verify if it works. Here's one of my sample nbs

```



import os
import asyncio
from datetime import datetime
from typing_extensions import TypedDict
from typing import Annotated
from operator import add

from langgraph.graph import StateGraph, START, END
from langchain_core.runnables import RunnableConfig

from galileo import galileo_context
from galileo.handlers.langchain import GalileoCallback

from galileo_core.schemas.logging.span import LlmSpan

import os
import time
from dotenv import load_dotenv

from langchain_core.agents import AgentAction, AgentFinish
from langchain_core.callbacks.base import BaseCallbackHandler
from langchain_core.documents import Document
from langchain_core.messages import BaseMessage
from langchain_core.outputs import LLMResult

from langchain_tavily import TavilySearch



LOCAL_API_KEY = ".."
os.environ["GALILEO_API_KEY"] = LOCAL_API_KEY
os.environ["GALILEO_PROJECT"]    = "pratyusha-demo-langgraph"
os.environ["GALILEO_CONSOLE_URL"] = "https://console-galileo-v2-dev.gcp-dev.galileo.ai"
os.environ["GALILEO_LOG_STREAM"] = "notebook-demo-2"
os.environ["TAVILY_API_KEY"] = ".."

galileo_context.init()

tavily_tool = TavilySearch(max_results=2)


# State for the orchestrator
class OrchestratorState(TypedDict):
    messages: Annotated[list[str], add]
    agent_choice: str
    user_query: str


# Supply Chain Agent
class SupplyChainState(TypedDict):
    messages: Annotated[list[str], add]

def supply_chain_news_node(state: SupplyChainState) -> SupplyChainState:
    """Node that gets latest supply chain news using Tavily tool"""
    query = "latest supply chain news and disruptions"
    result = tavily_tool.invoke(query)
    return {"messages": state["messages"] + [f"Supply Chain Agent: {result}"]}

# Build the supply chain agent as a StateGraph
supply_chain_graph = StateGraph(SupplyChainState)
supply_chain_graph.add_node("supply_chain_news", supply_chain_news_node)
supply_chain_graph.add_edge(START, "supply_chain_news")
supply_chain_graph.add_edge("supply_chain_news", END)
supply_chain_agent_app = supply_chain_graph.compile()







# Translator Agent
class TranslatorState(TypedDict):
    messages: Annotated[list[str], add]
    text_to_translate: str

def translator_node(state: TranslatorState) -> TranslatorState:
    """Node that translates English to Hindi"""
    text = state.get("text_to_translate", "Hello, how are you?")
    translation = f"Hindi translation of '{text}': नमस्ते, आप कैसे हैं?"
    return {"messages": state["messages"] + [translation], "text_to_translate": text}

# Build the translator agent as a StateGraph
translator_graph = StateGraph(TranslatorState)
translator_graph.add_node("translator", translator_node)
translator_graph.add_edge(START, "translator")
translator_graph.add_edge("translator", END)
translator_agent_app = translator_graph.compile()

# Orchestrator decision function
def orchestrator_decision(state: OrchestratorState) -> OrchestratorState:
    """Decides which agent to call based on user query"""
    query = state.get("user_query", "").lower()
    
    if any(keyword in query for keyword in ["supply chain", "news", "disruption", "logistics"]):
        return {"agent_choice": "supply_chain", "messages": state["messages"] + ["Orchestrator: Choosing Supply Chain Agent"]}
    elif any(keyword in query for keyword in ["translate", "hindi", "translation"]):
        return {"agent_choice": "translator", "messages": state["messages"] + ["Orchestrator: Choosing Translator Agent"]}
    else:
        # Default to supply chain for general queries
        return {"agent_choice": "supply_chain", "messages": state["messages"] + ["Orchestrator: Defaulting to Supply Chain Agent"]}

# Agent execution function
def execute_agent(state: OrchestratorState) -> OrchestratorState:
    """Executes the chosen agent"""
    agent_choice = state.get("agent_choice", "supply_chain")
    
    if agent_choice == "supply_chain":
        # Create supply chain agent state and execute the agent graph
        supply_chain_state = {"messages": state["messages"]}
        result = supply_chain_agent_app.invoke(supply_chain_state)
        return {"messages": result["messages"]}
    
    elif agent_choice == "translator":
        # Create translator agent state and execute the agent graph
        translator_state = {"messages": state["messages"], "text_to_translate": "Hello, how are you?"}
        result = translator_agent_app.invoke(translator_state)
        return {"messages": result["messages"]}
    
    return state


def news_agent_node(state: OrchestratorState) -> OrchestratorState:
    # Step 1: Get latest supply chain news in English
    query = "latest celebrity news"
    news_result = tavily_tool.invoke(query)
    english_message = f"Celebrity News: {news_result}"
    # Step 2: Translate the news to Hindi using the translator agent
    translator_state = {"messages": [english_message], "text_to_translate": news_result}
    translation_result = translator_agent_app.invoke(translator_state)
    # Step 3: Return both English and Hindi messages
    return {"messages": state["messages"] + [english_message] + translation_result["messages"]}

# Build the orchestrator graph
orchestrator_graph = StateGraph(OrchestratorState)

orchestrator_graph.add_node("decision", orchestrator_decision)
orchestrator_graph.add_node("execute_agent", execute_agent)
orchestrator_graph.add_node("news", news_agent_node)

orchestrator_graph.add_edge(START, "decision")
orchestrator_graph.add_edge("decision", "execute_agent")
orchestrator_graph.add_edge("execute_agent", "news")
orchestrator_graph.add_edge("news", END)

orchestrator_app = orchestrator_graph.compile()


'''
########################
# MAIN EXECUTION
########################
'''

def main():
    g_callback = GalileoCallback()
    config = RunnableConfig(callbacks=[g_callback])

    # Test the orchestrator with supply chain query
    print("\n=== Testing Orchestrator with Supply Chain Query ===")
    orchestrator_result = orchestrator_app.invoke({
        "messages": ["Starting orchestrator"],
        "user_query": "Get me the latest supply chain news"
    }, config=config)
    print("Orchestrator result:", orchestrator_result)

    galileo_context.flush()
    print("Session flushed and closed.")


if __name__ == "__main__":
    main()
    
```

rfriel:
> Approved! Please see if it's reasonable to 1) add tests to check for "name" in kwargs["metadata"]

Done

> and 2) run a few nbs (maybe sid2harthgupta can also run his langchain agent) and we can verify if it works. Here's one of my sample nbs

Sure. Here's two examples, with links to logstreams on dev (each logstream has only 1 session so I've linked to the session):

- The code you provided:
  - Using the current main branch: [link](https://console-galileo-v2-dev.gcp-dev.galileo.ai/project/158d5d9d-a480-477f-a842-bbcea4cd75df/log-streams/20f1791c-e5b6-4978-a68e-38cbc50f65ce/sessions/98a6d691-228a-425d-9b77-e5edf523c86d?drawerLevel=parameters&tabView=conversation&traceId=cbfebe4a-1083-40df-90c6-7ba63ebc155b)
  - Using this PR branch: [link](https://console-galileo-v2-dev.gcp-dev.galileo.ai/project/158d5d9d-a480-477f-a842-bbcea4cd75df/log-streams/6b744d15-7f65-4f87-96ee-0dd24b12cbbc/sessions/07cfe3ea-d128-4a23-8ce7-2660ffb87df5?drawerLevel=parameters&tabView=conversation&traceId=2bdc622c-0258-49be-82f2-d70e485b3678)
- [langgraph-fsi-agent](https://github.com/rungalileo/sdk-examples/tree/b6522dc6133876785e5313599d0f4352e1c905be/python/agent/langgraph-fsi-agent) from sdk-examples repo
  - Using the current main branch: [link](https://console-galileo-v2-dev.gcp-dev.galileo.ai/project/10732aeb-cafe-4ddb-a03d-ed069001b1f1/log-streams/10fe2f07-900d-4d63-864e-6015bb4ebccc/sessions/f23a4bbc-e996-4f43-a54c-50a82f5a5913?drawerLevel=parameters&tabView=conversation&traceId=6fc8802a-e5b1-4313-85e4-b4da60d3f71c)
  - Using this PR branch: [link](https://console-galileo-v2-dev.gcp-dev.galileo.ai/project/10732aeb-cafe-4ddb-a03d-ed069001b1f1/log-streams/1f89251f-b3fe-40df-ae45-2a2a8df5de4f/sessions/6017b1b0-c23d-4093-931a-c610a7a1da7e?drawerLevel=parameters&tabView=conversation&traceId=88414b78-a666-4806-b46e-ae005a4d11c4)

Everything looks good to me in the UI, lmk if you have any further questions

PR #250, opened 2025-07-24
fix: Rename rulesets to prioritized_rulesets and fix the type 

jrhone:
**Shortcut:** https://app.shortcut.com/galileo/story/36392/create-stage-and-update-stage-takes-a-list-of-rules-but-should-take-a-list-of-rulesets [sc-36392]

**Description:** `create_stage` and `update_stage` takes a list of rules, but should take a list of rulesets. They should both be a list of rulesets, named `prioritized_rulesets` to match the call to protect.

### DIFF
--- a/src/galileo/stages.py
+++ b/src/galileo/stages.py
@@ -1,3 +1,4 @@
+from collections.abc import Sequence
 from os import getenv
 from typing import Optional, Union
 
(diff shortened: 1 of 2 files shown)


jimbobbennett:
Interface looks good, tested locally and works for me.

ajaynayak (approved):

PR #254, opened 2025-07-28
feat: Use API client and config from `galileo-core`

jrhone:
**shortcut:** https://app.shortcut.com/galileo/story/34869/galileo-python-migrate-all-api-interactions-to-use-config-and-api-client-from-galileo-core#activity-36213 [sc-34869]

**description:** Migrate the api client to the Core config and client. We modify the openapi-autogen templates to use the Core client directly and use the Core config in our modules.

### DIFF
--- a/src/galileo/resources/api/auth/login_api_key_login_api_key_post.py
+++ b/src/galileo/resources/api/auth/login_api_key_login_api_key_post.py
@@ -3,8 +3,10 @@ from typing import Any, Optional, Union
 
 import httpx
 
+from galileo_core.constants.request_method import RequestMethod
+from galileo_core.helpers.api_client import ApiClient
+
 from ... import errors
-from ...client import AuthenticatedClient, Client
 from ...models.api_key_login_request import ApiKeyLoginRequest
 from ...models.http_validation_error import HTTPValidationError
 from ...models.token import Token
@@ -14,19 +16,17 @@ from ...types import Response
 def _get_kwargs(*, body: ApiKeyLoginRequest) -> dict[str, Any]:
     headers: dict[str, Any] = {}
 
-    _kwargs: dict[str, Any] = {"method": "post", "url": "/login/api_key"}
+    _kwargs: dict[str, Any] = {"method": RequestMethod.POST, "return_raw_response": True, "path": "/login/api_key"}
 
     _kwargs["json"] = body.to_dict()
 
     headers["Content-Type"] = "application/json"
 
-    _kwargs["headers"] = headers
+    _kwargs["content_headers"] = headers
     return _kwargs
 
 
-def _parse_response(
-    *, client: Union[AuthenticatedClient, Client], response: httpx.Response
-) -> Optional[Union[HTTPValidationError, Token]]:
+def _parse_response(*, client: ApiClient, response: httpx.Response) -> Optional[Union[HTTPValidationError, Token]]:
     if response.status_code == 200:
         response_200 = Token.from_dict(response.json())
 
@@ -41,9 +41,7 @@ def _parse_response(
         return None
 
 
-def _build_response(
-    *, client: Union[AuthenticatedClient, Client], response: httpx.Response
-) -> Response[Union[HTTPValidationError, Token]]:
+def _build_response(*, client: ApiClient, response: httpx.Response) -> Response[Union[HTTPValidationError, Token]]:
     return Response(
         status_code=HTTPStatus(response.status_code),
         content=response.content,
@@ -52,9 +50,7 @@ def _build_response(
     )
 
 
-def sync_detailed(
-    *, client: Union[AuthenticatedClient, Client], body: ApiKeyLoginRequest
-) -> Response[Union[HTTPValidationError, Token]]:
+def sync_detailed(*, client: ApiClient, body: ApiKeyLoginRequest) -> Response[Union[HTTPValidationError, Token]]:
     """Login Api Key
 
     Args:
@@ -70,14 +66,12 @@ def sync_detailed(
 
     kwargs = _get_kwargs(body=body)
 
-    response = client.get_httpx_client().request(**kwargs)
+    response = client.request(**kwargs)
 
     return _build_response(client=client, response=response)
 
 
-def sync(
-    *, client: Union[AuthenticatedClient, Client], body: ApiKeyLoginRequest
-) -> Optional[Union[HTTPValidationError, Token]]:
+def sync(*, client: ApiClient, body: ApiKeyLoginRequest) -> Optional[Union[HTTPValidationError, Token]]:
     """Login Api Key
 
     Args:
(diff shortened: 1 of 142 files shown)


openapi-client-config.yaml
```diff
@@ -1,4 +1,3 @@
 post_hooks:
-  - "rm ../../../src/galileo/resources/README.md"
```
jrhone, line 2: We remove this file in `scripts/auto-generate-api-client.sh`

src/galileo/utils/core_api_client.py
```diff
@@ -87,24 +53,10 @@ def _make_request(
         data: Optional[dict] = None,
         files: Optional[dict] = None,
         params: Optional[dict] = None,
-        json_request_only: bool = False,
```
jrhone, line 90: I didn't see this being used at all

src/galileo/utils/core_api_client.py
```diff
@@ -34,51 +28,23 @@ class GalileoCoreApiClient:
         The ID of the project.
     log_stream_id : Optional[str]
         The ID of the log stream.
-    base_url : Optional[str]
```
setu4993, line 37: Why do we still need _anything_ from this class? Why can't we just get rid of it? The stuff it's using can be (and should be) stored in `config`. And API interactions can just use that.

src/galileo/config.py
```diff
@@ -0,0 +1,25 @@
+# mypy: disable-error-code=syntax
+# We need to ignore syntax errors until https://github.com/python/mypy/issues/17535 is resolved.
+from typing import Any, Optional
+
+from galileo_core.schemas.base_config import GalileoConfig as BaseConfig
+
+
+class GalileoConfig(BaseConfig):
+    # Config file for this project.
+    config_filename: str = "galileo-config.json"
```
setu4993, line 10:
```suggestion
    config_filename: str = "galileo-python-config.json"
```

src/galileo/config.py
```diff
@@ -0,0 +1,25 @@
+# mypy: disable-error-code=syntax
+# We need to ignore syntax errors until https://github.com/python/mypy/issues/17535 is resolved.
+from typing import Any, Optional
+
+from galileo_core.schemas.base_config import GalileoConfig as BaseConfig
+
+
+class GalileoConfig(BaseConfig):
```
setu4993, line 8: I'd strongly recommend renaming this to avoid confusion.
```suggestion
class GalileoPythonConfig(BaseConfig):
```

tests/conftest.py
```diff
@@ -1,20 +1,72 @@
...
+) -> Generator[None, None, None]:
+    config = GalileoConfig.get()
+    yield
+    config.reset()
+
+
 @pytest.fixture(autouse=True)
 def env_setup(monkeypatch):
     monkeypatch.setenv("OPENAI_API_KEY", "local")
+    monkeypatch.setenv("GALILEO_API_KEY", "api-1234567890")
```
setu4993, line 69: We shouldn't need these, no? They are only for authenticated requests.

Either way, if we need env vars for _all_ tests, they should be in `pyproject.toml` under the `pytest` config, like this: https://github.com/rungalileo/api/blob/main/pyproject.toml#L105-L141

src/galileo/base.py
```diff
@@ -9,13 +10,20 @@ class BaseClientModel:
...
         else:
-            self.client = GalileoApiClient()
+            self.config = GalileoConfig.get()
+
+    @property
+    def client(self) -> ApiClient:
+        """
+        Returns a fresh client from the config, ensuring the token is up-to-date.
+        """
+        return self.config.api_client
```
setu4993, line 29: Nice.

src/galileo/base.py
```diff
@@ -9,13 +10,20 @@ class BaseClientModel:
     Used by the model classes to lazy load the API client.
 
     Args:
-        client (Optional[GalileoApiClient], optional): The client to use. Defaults to None.
+        config (Optional[GalileoConfig], optional): The config to use. Defaults to None.
```
setu4993, line 13: Also curious why we need this class... This just looks like a silly wrapper on top of the config class?

setu4993 (commented):
Mostly looks good. 2 things stand out to be, which both seem like us not getting rid of some of the cruft with `BaseClientModel` and another class.

src/galileo/utils/core_api_client.py
```diff
@@ -87,24 +53,10 @@ def _make_request(
         data: Optional[dict] = None,
         files: Optional[dict] = None,
         params: Optional[dict] = None,
-        json_request_only: bool = False,
     ) -> Any:
-        if json_request_only:
-            content_headers = HttpHeaders.accept_json()
-        else:
-            content_headers = HttpHeaders.json()
-        headers = {**self.auth_header, **content_headers, **self.client_type_header}
```
jrhone, line 96: I think I need to add back in these headers except for the auth header

src/galileo/base.py
jrhone: That's exactly what it is : ) I subclassed it based on our brief conversation in slack, but I'm not actually using it to maintain any state.. it was more of a, "for the future", kinda thing.

I can remove and just use the core Config class directly.. or I can put some state in there, but I haven't looked properly into how the other repos are using their custom config classes to maintain/use the state.

Screenshot 2025-07-28 at 7 49 24 PM

setu4993: I was actually suggesting us using `GalileoConfig` directly / sub-classing it in here (which exists in the other file) but not using that in the init for this... Right now, we have a 3-layer thing: `BaseClient` => `Config` => `APIClient`, which seems a few too many abstractions.

setu4993 (approved):
Looks good for now. Would really be nice to get rid of some of the wrapper classes here though.

src/galileo/base.py
jrhone: Oh I see.. you mean `BaseClientModel` is the silly wrapper.. yea, I tried changing as little as possible to be a bit more expeditious, which left a bit of cruft in there for sure. 

I'll circle back on this one and add a ticket to address this in the near future.  

Slightly under the weather since yesterday so I'm a bit cloudy and the math just ain't math'ing in my head at the mo.

setu4993: Sounds good, thanks Jonathan.

src/galileo/utils/core_api_client.py
jrhone: So the python client had two ways of making api requests.. 
- this class, used by the Logger, which has calls for traces, spans and sessions. It used the Core Client under the hood, but not Core Config.
- the generated api resources, used by everything else

I tried being as surgical with my changes as possible to get things done quickly, with a very low chance of breaking anything.. but definitely this class should be gotten rid of. (I only replaced it's direct usage of Core Client with Core Client from Config)

Ideally tho I think the Logger should use the generated api resources like the rest of the library, which is a slightly larger lift than I think was planned for this ticket, so I added this to the follow up [streamline ticket](https://app.shortcut.com/galileo/story/36664/streamline-the-usage-of-core-config-in-galileo-python).

src/galileo/base.py
jrhone: Wlc

PR #263, opened 2025-08-07
poc

pratduv:
TL;DR: rn we have logic for `flush()`, `update_trace/step_streaming()` etc that calls a sync/async function in core_api_client.py that has the same logic duplicated across the sync/async functions
Depending on whether it's a sync or async function, we then call `galileo-core api_client.py`  `request()` or `arequest()` .. but both eventually use the same `ApiClient.make_request` call
I think this balloons that file up for no reason  (unless I'm missing something important) and we should have async_run used instead at the logger.py level.

### DIFF
--- a/src/galileo/decorator.py
+++ b/src/galileo/decorator.py
@@ -930,6 +930,24 @@ class GalileoDecorator:
             name=name, previous_session_id=previous_session_id, external_id=external_id
         )
 
+    async def async_start_session(
+        self, name: Optional[str] = None, previous_session_id: Optional[str] = None, external_id: Optional[str] = None
+    ) -> str:
+        """
+        Async start a session in the active context logger instance.
+
+        Args:
+            name: The name of the session. If not provided, a session name will be generated automatically.
+            previous_session_id: The id of the previous session. Defaults to None.
+            external_id: The external id of the session. Defaults to None.
+
+        Returns:
+            str: The id of the newly created session.
+        """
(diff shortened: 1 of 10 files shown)


src/galileo/utils/core_api_client.py
```diff
@@ -85,18 +85,6 @@ async def ingest_traces(self, traces_ingest_request: TracesIngestRequest) -> dic
...
-        elif self.log_stream_id:
-            traces_ingest_request.log_stream_id = UUID(self.log_stream_id)
-
-        json = traces_ingest_request.model_dump(mode="json")
-
-        return self._make_request(
-            RequestMethod.POST, endpoint=Routes.traces.format(project_id=self.project_id), json=json
-        )
-
     async def ingest_spans(self, spans_ingest_request: SpansIngestRequest) -> dict[str, str]:
```
pratduv, line 88: I want to do what I did with `ingest_traces` to all other functions in this file. Do we really need both `_make_request` and _make_async_request` when under the hood request calls `async_run` in the end (via `galileo-core`? We have a lot of logic duplication in this file

src/galileo/decorator.py
```diff
@@ -930,6 +930,24 @@ def start_session(
             name=name, previous_session_id=previous_session_id, external_id=external_id
         )
 
+    async def async_start_session(
```
pratduv, line 933: ignore this change for now

PR #268, opened 2025-08-08
chore: move Protect models to galileo.protect to avoid conflicts

ajaynayak:
Also including a few more Protect models


src/galileo/__init__.py
```diff
@@ -21,11 +21,5 @@
 from galileo_core.schemas.logging.span import LlmSpan, RetrieverSpan, Span, StepWithChildSpans, ToolSpan, WorkflowSpan
 from galileo_core.schemas.logging.step import StepType
 from galileo_core.schemas.logging.trace import Trace
-from galileo_core.schemas.protect.execution_status import ExecutionStatus
-from galileo_core.schemas.protect.payload import Payload
-from galileo_core.schemas.protect.request import Request
-from galileo_core.schemas.protect.response import Response
-from galileo_core.schemas.protect.ruleset import Ruleset
-from galileo_core.schemas.protect.stage import StageType
```
setu4993, line 29: We should probably export all of these from here as `ProtectPayload`, `ProtectRequest`, etc.

PR #270, opened 2025-08-08
feat: Merge sync/async functions in core api client

pratduv:


pratduv:
https://app.shortcut.com/galileo/story/34836/galileo-python-add-async-support-to-all-log-stream-ingest-methods

PR #288, opened 2025-08-19
feat: Streamline Config Client usage by removing Base Client Model

jrhone:
**shortcut:** https://app.shortcut.com/galileo/story/36664/streamline-the-usage-of-core-config-in-galileo-python [sc-36664]

**description:** BaseClientModel -> Config -> Client is too many layers of abstraction so we remove BaseClientModel and get a Config from the Singleton in the class or method directly.


src/galileo/datasets.py
```diff
@@ -38,12 +37,13 @@ class DatasetAPIException(APIException):
     pass
 
 
-class Dataset(BaseClientModel, DecorateAllMethods):
+class Dataset(DecorateAllMethods):
     content: Optional[DatasetContent] = None
+    config: GalileoPythonConfig
 
-    def __init__(self, dataset_db: DatasetDB, config: Optional[GalileoPythonConfig] = None) -> None:
```
vamaq, line 44: I like this simplification, my only concern is changing the signature of the constructor and removing the possibility to pass the configuration as a parameter. Did we validate that we don't really want to allow that option?

PR #297, opened 2025-08-27
39509: Fix info message when flushing traces

sid2harthgupta:
Based on recent customer feedback (Ben Dechrai), flushing a single trace currently says "Successfully flushed 1 traces" instead of the singular "trace".

https://app.shortcut.com/galileo/story/39509/flushing-a-single-trace-says-successfully-flushed-1-traces-instead-of-the-singular-trace


sid2harthgupta:
> Please add a test here
> 
> Is self.traces = None a possibility? In which case the whole thing could crash

@nachiket-galileo This PR does not change any of the logic. `logged_traces = self.traces` was already in use and this just moves it a few lines up. The only change is to the text content of the log messages which should not require any new tests.

PR #307, opened 2025-09-04
chore: Adding support for loading  projects from env vars in experiments

jimbobbennett:
The `run_experiment` function expects a project name. To make this more consistent with other functions, this change adds the ability to:

- Pass in a project by Id (the other experiment functions only take the Id)
- Not pass in a project and have it loaded from an environment variable

[SC-40118](https://app.shortcut.com/galileo/story/40118/run-experiment-should-support-project-ids-and-projects-from-env-vars)

### DIFF
--- a/src/galileo/experiments.py
+++ b/src/galileo/experiments.py
@@ -1,6 +1,7 @@
 import builtins
 import datetime
 import logging
+import os
 from typing import Any, Callable, Optional, Union
 
 from attrs import define as _attrs_define
@@ -233,6 +234,7 @@ def run_experiment(
     prompt_template: Optional[PromptTemplate] = None,
     prompt_settings: Optional[PromptRunSettings] = None,
     project: Optional[str] = None,
+    project_id: Optional[str] = None,
     dataset: Optional[Union[Dataset, list[dict[str, str]], str]] = None,
     dataset_id: Optional[str] = None,
     dataset_name: Optional[str] = None,
@@ -252,7 +254,8 @@ def run_experiment(
         experiment_name: Name of the experiment
         prompt_template: Template for prompts
         prompt_settings: Settings for prompt runs
-        project: Project name
+        project: Project name. If neither project name nor project_id is provided, the GALILEO_PROJECT environment variable will be used
+        project_id: Project Id. If neither project name nor project_id is provided, the GALILEO_PROJECT environment variable will be used
         dataset: Dataset object, list of records, or dataset name
         dataset_id: ID of the dataset
         dataset_name: Name of the dataset
@@ -265,9 +268,26 @@ def run_experiment(
     Raises:
         ValueError: If required parameters are missing or invalid
     """
-    # Get project
-    if project is None:
-        raise ValueError("A project name must be provided")
+    # Get the project
+    # If the name or Id is set, then use these to get the project. Only one can be provided
+    # If neither are set, use the environment variables to get the project name or Id
+    project_name = project or os.getenv("GALILEO_PROJECT_NAME", os.getenv("GALILEO_PROJECT"))
+    project_id = project_id or os.getenv("GALILEO_PROJECT_ID")
+
+    # Check we only have one of project name or Id.
+    if project_name is None and project_id is None:
+        raise ValueError("A project name or Id must be provided")
+    if project_name is not None and project_id is not None:
+        raise ValueError("Only one of project name or Id should be provided")
+
+    # Get the project from the name or Id
+    project_obj = Projects().get(id=project_id) if project_id else Projects().get(name=project_name)
+
+    # Ensure we have a valid project
+    if not project_obj:
+        if project_id:
+            raise ValueError(f"Project with Id {project_id} does not exist")
+        raise ValueError(f"Project {project_name} does not exist")
 
     # Load dataset and records
     dataset_obj, records = load_dataset_and_records(dataset, dataset_id, dataset_name)
@@ -283,11 +303,6 @@ def run_experiment(
 
     if function and prompt_template:
         raise ValueError("A function or prompt_template should be provided, but not both")
-
-    # Get project
-    project_obj = Projects().get(name=project)
-    if not project_obj:
-        raise ValueError(f"Project {project} does not exist")
 
     # Create or get experiment
     existing_experiment = Experiments().get(project_obj.id, experiment_name)
--- a/src/galileo/experiments.py
+++ b/src/galileo/experiments.py
@@ -268,6 +268,21 @@ def run_experiment(
     Raises:
         ValueError: If required parameters are missing or invalid
     """
+    # Load dataset and records
+    dataset_obj, records = load_dataset_and_records(dataset, dataset_id, dataset_name)
+
+    # Validate experiment configuration
+    if prompt_template and not dataset_obj:
+        raise ValueError("A dataset record, id, or name of a dataset must be provided when a prompt_template is used")
+
+    if function and not records:
+        raise ValueError(
+            "A dataset record, id or name of a dataset, or list of records must be provided when a function is used"
+        )
+
+    if function and prompt_template:
+        raise ValueError("A function or prompt_template should be provided, but not both")
+
     # Get the project
     # If the name or Id is set, then use these to get the project. Only one can be provided
     # If neither are set, use the environment variables to get the project name or Id
@@ -288,21 +303,6 @@ def run_experiment(
         if project_id:
             raise ValueError(f"Project with Id {project_id} does not exist")
         raise ValueError(f"Project {project_name} does not exist")
-
-    # Load dataset and records
-    dataset_obj, records = load_dataset_and_records(dataset, dataset_id, dataset_name)
-
-    # Validate experiment configuration
-    if prompt_template and not dataset_obj:
-        raise ValueError("A dataset record, id, or name of a dataset must be provided when a prompt_template is used")
-
-    if function and not records:
-        raise ValueError(
-            "A dataset record, id or name of a dataset, or list of records must be provided when a function is used"
-        )
-
-    if function and prompt_template:
-        raise ValueError("A function or prompt_template should be provided, but not both")
 
     # Create or get experiment
     existing_experiment = Experiments().get(project_obj.id, experiment_name)
--- a/src/galileo/experiments.py
+++ b/src/galileo/experiments.py
@@ -228,6 +228,46 @@ def process_row(row: DatasetRecord, process_func: Callable) -> str:
     return output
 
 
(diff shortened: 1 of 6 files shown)


src/galileo/experiments.py
```diff
@@ -284,10 +283,33 @@ def run_experiment(
...
+
+    # Get the project from the name or Id
+    project_obj = (
+        Projects().get(id=project_id) if project_id else (Projects().get(name=project_name) if project_name else None)
+    )
+
+    # Ensure we have a valid project
     if not project_obj:
-        raise ValueError(f"Project {project} does not exist")
+        if project_id:
```
sid2harthgupta, line 310: Is this meant to be `if not project_id`?

src/galileo/experiments.py
```diff
@@ -284,10 +283,33 @@ def run_experiment(
...
 
-    # Get project
-    project_obj = Projects().get(name=project)
+    # Get the project
+    # If the name or Id is set, then use these to get the project. Only one can be provided
+    # If neither are set, use the environment variables to get the project name or Id
+    project_name = project or os.getenv("GALILEO_PROJECT")
+    project_id = project_id or os.getenv("GALILEO_PROJECT_ID")
+
+    if project_name == "":
```
sid2harthgupta, line 292: Should we do `if project_name.strip() == ""` and below as well?

src/galileo/experiments.py
```diff
@@ -284,10 +283,33 @@ def run_experiment(
     if function and prompt_template:
         raise ValueError("A function or prompt_template should be provided, but not both")
 
-    # Get project
-    project_obj = Projects().get(name=project)
+    # Get the project
+    # If the name or Id is set, then use these to get the project. Only one can be provided
+    # If neither are set, use the environment variables to get the project name or Id
+    project_name = project or os.getenv("GALILEO_PROJECT")
+    project_id = project_id or os.getenv("GALILEO_PROJECT_ID")
```
sid2harthgupta, line 290: Shouldn't these lines be below the checks below, so just in case the project/project_id is an empty string, the environment variables are still read?

src/galileo/experiments.py
```diff
@@ -284,10 +283,33 @@ def run_experiment(
...
+    # Check we only have one of project name or Id.
+    if project_name is None and project_id is None:
+        raise ValueError("A project name or Id must be provided")
+    if project_name is not None and project_id is not None:
+        raise ValueError("Only one of project name or Id should be provided")
+
+    # Get the project from the name or Id
+    project_obj = (
+        Projects().get(id=project_id) if project_id else (Projects().get(name=project_name) if project_name else None)
+    )
```
sid2harthgupta, line 306: Wdyt about abstracting the logic for retrieving the project object and error handling into another function so it can be reused at all places?

src/galileo/experiments.py
```diff
@@ -284,10 +283,33 @@ def run_experiment(
...
+
+    # Check we only have one of project name or Id.
+    if project_name is None and project_id is None:
+        raise ValueError("A project name or Id must be provided")
+    if project_name is not None and project_id is not None:
+        raise ValueError("Only one of project name or Id should be provided")
+
+    # Get the project from the name or Id
+    project_obj = (
+        Projects().get(id=project_id) if project_id else (Projects().get(name=project_name) if project_name else None)
```
sid2harthgupta, line 305: The final `else None` should never happen here, right? If not, let's drop it

sid2harthgupta (commented):
Added some comments. Happy to chat if helpful

src/galileo/experiments.py
jimbobbennett: Done

src/galileo/experiments.py
jimbobbennett: 100% agree, but this feels like a bigger change that impacts the entire SDK as we need to decide where to put it, then review all the code to remove duplication (e.g. the Logger code has some of this internally) and use the new way to do it. Then ensure going forward this code is always used.

How do you want to proceed on this? Raise a ticket for the team to look at?

src/galileo/experiments.py
```diff
@@ -227,12 +228,53 @@ def process_row(row: DatasetRecord, process_func: Callable) -> str:
...
+
+    # Check we have one and only one of project name or Id.
+    if not project_name and not project_id:
+        raise ValueError("A project name or Id must be provided")
+    if project_name and project_id:
+        raise ValueError("Only one of project name or Id should be provided")
+
+    # Get the project from the name or Id
+    project = (
+        Projects().get(id=project_id) if project_id else (Projects().get(name=project_name) if project_name else None)
```
sid2harthgupta, line 258: Can we use this get call?
https://github.com/rungalileo/galileo-python/blob/main/src/galileo/projects.py#L127

src/galileo/experiments.py
```diff
@@ -227,12 +228,53 @@ def process_row(row: DatasetRecord, process_func: Callable) -> str:
...
+    4. If the project name and Id are not passed, and both GALILEO_PROJECT and GALILEO_PROJECT_ID are provided, raise an error.
+    5. Use the provided or environment variable value to load the project. If the project does not exist, raise an error.
+
+    Args:
+        project_name: Name of the project
+        project_id: Id of the project
+    """
+    # Get the project name or Id from parameters or environment variables, stripping whitespace and converting empty strings to None
+    project_name = (project_name or os.getenv("GALILEO_PROJECT") or "").strip() or None
+    project_id = (project_id or os.getenv("GALILEO_PROJECT_ID") or "").strip() or None
```
sid2harthgupta, line 248: I don't think this works as intended:
```python
>>> "" or "hello"
'hello'
>>> " " or "hello"
' '
>>> " ".strip() or "hello"
'hello'
```
We could just convert it to:
```python
project_id or os.getenv("GALILEO_PROJECT_ID")
```
`os.getenv` will automatically return None when the environment variable does not exist

src/galileo/projects.py
```diff
@@ -149,9 +146,15 @@ def get(self, *, id: Optional[str] = None, name: Optional[str] = None) -> Option
             If the request takes longer than Client.timeout.
 
         """
-        if (id is None) and (name is None):
+        name = (name or os.getenv("GALILEO_PROJECT") or "").strip() or None
+        id = (id or os.getenv("GALILEO_PROJECT_ID") or "").strip() or None
+
+        # Check we have one and only one of project name or Id.
+        if (not name and not id) or (name and id):
```
sid2harthgupta, line 153: 👍

src/galileo/projects.py
```diff
@@ -149,9 +146,15 @@ def get(self, *, id: Optional[str] = None, name: Optional[str] = None) -> Option
...
             raise ValueError("Exactly one of 'id' or 'name' must be provided")
 
+        project: Optional[Project] = None
```
sid2harthgupta, line 156: Needed?

src/galileo/projects.py
```diff
@@ -149,9 +146,15 @@ def get(self, *, id: Optional[str] = None, name: Optional[str] = None) -> Option
             If the request takes longer than Client.timeout.
 
         """
-        if (id is None) and (name is None):
+        name = (name or os.getenv("GALILEO_PROJECT") or "").strip() or None
+        id = (id or os.getenv("GALILEO_PROJECT_ID") or "").strip() or None
```
sid2harthgupta, line 150: I don't think this belongs here- the decision to use the environment variables and the names of those variables should be decided by the callers

src/galileo/experiments.py
```diff
@@ -248,11 +249,16 @@ def run_experiment(
 
     When using a runner function, you can also pass a list of dictionaries to the function to act as a dataset.
 
+    To define the project to use, either set the project or project_id parameters, or set the GALILEO_PROJECT or GALILEO_PROJECT_ID.
```
sid2harthgupta, line 252: nit-
`or set the GALILEO_PROJECT or GALILEO_PROJECT_ID.`
to
`or set the GALILEO_PROJECT or GALILEO_PROJECT_ID environment variables.`

sid2harthgupta: In general this is sounding a bit redundant. Maybe rephrase to:
`The project can be specified by providing exactly one of the project name (via the 'project' parameter or the GALILEO_PROJECT environment variable) or the project ID (via the 'project_id' parameter or the GALILEO_PROJECT_ID environment variable).`

src/galileo/experiments.py
```diff
@@ -248,11 +249,16 @@ def run_experiment(
...
+    If you pass the project in the parameters, then only one of project or project_id should be provided. Setting both with give an error.
+    If you don't set either the project or project_id parameters, then only one of GALILEO_PROJECT or GALILEO_PROJECT_ID should be set in the environment variables. Setting both with give an error.
+
     Args:
         experiment_name: Name of the experiment
         prompt_template: Template for prompts
         prompt_settings: Settings for prompt runs
-        project: Project name
+        project: Project name. Pass either only one of project or project_id, or pass nothing and the GALILEO_PROJECT or GALILEO_PROJECT_ID environment variables will be used.
```
sid2harthgupta, line 260: Simplify to something like-
`project: Optional project name. Takes preference over the GALILEO_PROJECT environment variable. Leave empty if using project_id`
and same for project_id

src/galileo/experiments.py
```diff
@@ -284,9 +286,13 @@ def run_experiment(
     if function and prompt_template:
         raise ValueError("A function or prompt_template should be provided, but not both")
 
-    # Get project
-    project_obj = Projects().get(name=project)
+    # Get the project from the name or Id
+    project_obj = Projects().get(id=project_id, name=project)
```
sid2harthgupta, line 290: This could become:
```python
project_obj = Projects().get(
    id=project_id.strip() or os.getenv("GALILEO_PROJECT"), 
    name=project.strip() or os.getenv("GALILEO_PROJECT_ID")
)
```

sid2harthgupta (commented):
A few more thoughts but almost there

src/galileo/stages.py
```diff
@@ -31,19 +30,14 @@ def _get_project_id(
...
-    project_name = project_name if project_name else getenv("GALILEO_PROJECT")
-
-    if project_name:
-        project = Projects(config=config).get(name=project_name)
-        if not project:
-            raise ValueError(f"Project with name '{project_name}' not found.")
-        return str(project.id)
-
-    raise ValueError("Either project_id or project_name must be provided.")
+    # If the project Id is a UUID4, convert to string.
```
sid2harthgupta, line 33: Unrelated to this PR- it is confusing why `_get_project_id` accepts a `project_id` in the first place

sid2harthgupta (approved):
LGTM 👍

PR #321, opened 2025-09-16
feat: Enable metrics for LogStream

nachiket-galileo:
Shortcut: [sc-39869 / [Python SDK] Improve the way metrics are added to log traces](https://app.shortcut.com/galileo/story/39869/python-sdk-improve-the-way-metrics-are-added-to-log-traces)


vamaq (commented):
When trying to run the example I got the following error:

```python
>>> from galileo.log_streams import enable_metrics  # noqa
>>> from galileo.schema.metrics import GalileoScorers # noqa
>>> scorer_configs, local_metrics = enable_metrics(
...     log_stream_name=log_stream_name,
...     project_name=project_name,
...     metrics=[
...         GalileoScorers.correctness,
...         GalileoScorers.completeness,
...         "context_relevance",
...         "instruction_adherence"
...     ]
... )
Error occurred during execution: create_metric_configs: LogStreams.create_metric_configs() takes 3 positional arguments but 4 were given
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: cannot unpack non-iterable NoneType object
```

nachiket-galileo:
@john-weiler not understanding the context for this ticket here?

john-weiler:
> @john-weiler not understanding the context for this ticket here?

Sorry. Yeah for the UI we have some checks that check if certain metrics are not able to be enabled at log stream level (due to needing ground truth). So to prevent regression here, we should add it to the UI

src/galileo/utils/metrics.py
```diff
@@ -28,3 +31,82 @@ def _populate_local_metric(
         metric_value = local_metric.scorer_fn(step)
         setattr(step.metrics, local_metric.name, local_metric.scorer_fn(step))
         scores.append(metric_value)
+
+
+def create_metric_configs(
```
nachiket-galileo, line 36: @vamaq @john-weiler 

Consolidating the create_metric_configs here so that experiments and log_streams can both use the same

src/galileo/log_streams.py
```diff
@@ -92,6 +118,110 @@ def __init__(self, log_stream: Union[None, LogStreamResponse] = None):
             self.additional_properties = log_stream.additional_properties.copy()
             return
 
+    def enable_metrics(
+        self, metrics: builtins.list[Union[GalileoScorers, Metric, LocalMetricConfig, str]]
+    ) -> builtins.list[LocalMetricConfig]:
+        """
+        Enable metrics directly on this log stream instance.
```
vamaq, line 125: I like documentation, but this feels too verbose. I’d remove subjective wording like `most intuitive`, `is ideal`, or `making it perfect for all use cases`.

For `Parameters`, `Returns`, and `Raises`, I think it’s fine to define them in one method and reference that validation in another docs-tring. This avoids duplication, makes it easier to maintain, and helps users understand the relationships between methods.

Tegarding the `Examples`, I’d either remove them from the Service methods entirely or keep just one.

cc  @jimbobbennett

vamaq (approved):
LGTM!

PR #322, opened 2025-09-17
feat: Update LangChain Handler to Support Responses API

mikebranc:
### Summary
Adds support for the responses API in LangChain Callback. Two things are required to add support for LangChain.
### Changes
1. Handle support for new `content` format in `response` in the handler's `on_llm_end` method
    1. The responses API returns content as a list of dictionaries with the following format: `content: { 'type': 'text', 'text': 'some text', 'annotations': [] }`. This differs from content returned from the completions API which is a string.
    2. Without this change, the output saved to galileo is a list, causing the UI to break
2. Add support for reasoning
    1. the responses API allows users to return reasoning  from the LLM
    2. This PR allows the user to save reasoning on the message as `additional_kwargs` which follows the same pattern used by OpenInference

Langchain Responses API docs: https://docs.langchain.com/oss/python/integrations/chat/openai#responses-api

### Ticket
[SC-41100](https://app.shortcut.com/galileo/story/41100/support-openai-responses-api-in-langchain-callback)


Focadecombate (approved):
👍

PR #330, opened 2025-09-24
feat: Upgrade linters for ruff and mypy

nachiket-galileo:
Shortcut: https://app.shortcut.com/galileo/story/41545/python-sdk-add-linter-to-build-process

We aim to add much stronger and stricter linters to existing mypy and ruff. 
A few considerations

- We ignore resources for now since these are auto-gen
- We are more lenient currently on tests but that will change

### DIFF
--- a/src/galileo/job_progress.py
+++ b/src/galileo/job_progress.py
@@ -56,9 +56,9 @@ def scorer_jobs_status(project_id: str, run_id: str) -> None:
     scorer_jobs = get_run_scorer_jobs(project_id, run_id)
     for job in scorer_jobs:
         scorer_name = None
-        if job.request_data.get("prompt_scorer_settings"):
+        if "prompt_scorer_settings" in job.request_data:
             scorer_name = job.request_data["prompt_scorer_settings"]["scorer_name"]
-        elif job.request_data.get("scorer_config"):
+        elif "scorer_config" in job.request_data:
             scorer_name = job.request_data["scorer_config"]["name"]
 
         if not scorer_name:


src/galileo/handlers/openai_agents/handler.py
```diff
@@ -324,18 +324,8 @@ def on_span_end(self, span: Span[Any]) -> None:
...
-                    **wf_data,
-                    "metadata": {**node.span_params.get("metadata", {}), **wf_data.get("metadata", {})},
-                    "status_code": wf_data.get("status_code", node.span_params.get("status_code", 200)),
-                }
-            )
-            # Workflow output might only be known at the end
-            if node.span_params.get("output") is None:
-                node.span_params["output"] = wf_data.get("output")
+        # Note: workflow type is not currently supported in the node_type literal
+        # elif galileo_type == "workflow": (removed as unreachable code)
```
vamaq, line 328: 👍

vamaq (approved):
Nice!

nachiket-galileo:
> yes removing this:
> 
> ```
>  elif galileo_type == "workflow":
>             wf_data = _extract_workflow_data(span.span_data)
>             end_params.update(
>                 {
>                     **wf_data,
>                     "metadata": {**node.span_params.get("metadata", {}), **wf_data.get("metadata", {})},
>                     "status_code": wf_data.get("status_code", node.span_params.get("status_code", 200)),
>                 }
>             )
>             # Workflow output might only be known at the end
>             if node.span_params.get("output") is None:
>                 node.span_params["output"] = wf_data.get("output")
> ```
> 
> is a breaking change

@franz101 I dont think this is a breaking change. What am I missing here? https://github.com/rungalileo/galileo-python/blob/main/src/galileo/schema/handlers.py#L11

PR #331, opened 2025-09-25
chore: Update OpenAPI Spec for SDK

nachiket-galileo:


jrhone (changes requested):
Let's check if openapi modified their `codegen_templates/endpoint_module.py.jinja` file and if we need to pull those changes into our version and re-apply our modifications to that file back on top :)

src/galileo/resources/api/datasets/bulk_delete_datasets_datasets_bulk_delete_delete.py
```diff
@@ -26,6 +27,8 @@ def _get_kwargs(*, body: BulkDeleteDatasetsRequest) -> dict[str, Any]:
 
     headers["Content-Type"] = "application/json"
 
+    headers["X-Galileo-SDK"] = f"galileo-python/{get_package_version()}"
```
jrhone, line 30: Did you add this manually?

src/galileo/resources/api/datasets/bulk_delete_datasets_datasets_bulk_delete_delete.py
```diff
@@ -34,17 +37,14 @@ def _parse_response(
     *, client: ApiClient, response: httpx.Response
 ) -> Optional[Union[BulkDeleteDatasetsResponse, HTTPValidationError]]:
     if response.status_code == 200:
-        response_200 = BulkDeleteDatasetsResponse.from_dict(response.json())
```
jrhone, line 37: So it generates with `if response.status_code == :`
Then the script runs to convert to `if response.status_code == 200:`

Then the linter makes this change?

jrhone:
> > Let's check if openapi modified their `codegen_templates/endpoint_module.py.jinja` file and if we need to pull those changes into our version and re-apply our modifications to that file back on top :)
> 
> @jrhone Quick FYI the updates dont work and are still breaking for me

Gotcha... I'm not sure about using a script to fix what should be built-in functionality.

I mean we're not customizing the functionality.. the tool is just breaking.

I suppose if no other fix then happy to approve, but, what are your thoughts on that?

jrhone:
Wasn't possible to separate linter changes from actual changes?

src/galileo/resources/api/datasets/bulk_delete_datasets_datasets_bulk_delete_delete.py
jrhone: Ok I see it's in the jinja template.. interesting that it was missing from this file and not the rest, I guess doesn't matter 🤷🏾‍♂️

jrhone:
Should we pull in the changes from the new templates since we updated our openapi-generators version?

src/galileo/resources/api/datasets/bulk_delete_datasets_datasets_bulk_delete_delete.py
jrhone: But actually yeah I haven't checked every file.. github is forcing me to look at each file one at a time 😂

jrhone (approved):
Ok I guess my only real comment above is what happens if we update our `openapi-python-client` to the latest.. maybe it will fix the status code issue since there were changes directly related to status codes in the new version.

Ideally we wouldn't fix breaking functionality with a patch.. but otherwise looks good.

PR #333, opened 2025-09-26
fix: Custom metric creation ignores some parameters

jrhone:
**shortcut:** https://app.shortcut.com/galileo/story/40922/python-sdk-custom-metric-creation-ignores-some-parameters [sc-40922]

**description:** Custom metric creation ignores some parameters because they were sent in the create version request instead of the create scorer request.

This might be a bug with the API or a bug in the UI but instead **this PR aligns the SDK with the UI in terms of how the API is called**, which currently works.

I'll file a separate ticket to investigate why the API isn't working when sending the data in the create version request only and not as defaults in the create metric request.

### DIFF
--- a/src/galileo/scorers.py
+++ b/src/galileo/scorers.py
@@ -12,6 +12,7 @@ from galileo.resources.api.run_scorer_settings import (
 from galileo.resources.models import (
     HTTPValidationError,
     ListScorersRequest,
+    ListScorersResponse,
     ScorerConfig,
     ScorerResponse,
     ScorerTypeFilter,
@@ -30,7 +31,7 @@ class Scorers:
     def __init__(self) -> None:
         self.config = GalileoPythonConfig.get()
 
-    def list(self, types: Optional[list[ScorerTypes]] = None) -> Union[Unset, list[ScorerResponse]]:
+    def list(self, types: Optional[list[ScorerTypes]] = None) -> list[ScorerResponse]:
         """
         Args:
             types: List of scorer types to filter by. Defaults to all scorers.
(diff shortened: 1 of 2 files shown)


john-weiler (commented):
I’m not sure this is right - we want these values to be set at the version level.

jrhone:
> I’m not sure this is right - we want these values to be set at the version level.

Would love to discuss this further as I have no idea how it's meant to be working.. both the scorer and the version accept the same arguments..

In the UI however (at least when creating the metric initially), the data is only sent to create the scorer and the version request only receives the `user_prompt`.

nachiket-galileo (approved):

tests/test_metrics.py
```diff
@@ -118,11 +121,6 @@ def test_create_custom_llm_metric_success(
 
         assert isinstance(version_request, CreateLLMScorerVersionRequest)
         assert version_request.user_prompt == "Rate the quality of this response"
-        assert version_request.scoreable_node_types == [StepType.llm]
-        assert version_request.cot_enabled is True
-        assert version_request.output_type == OutputTypeEnum.BOOLEAN
```
nachiket-galileo, line 123: I am a bit confused about this change. Why are we removing all this?

src/galileo/metrics.py
```diff
@@ -65,20 +65,18 @@ def create_custom_llm_metric(
...
+                output_type=output_type,
+                cot_enabled=cot_enabled,
+                scoreable_node_types=[node_level],
+            ),
         )
 
         scorer = create_scorers_post.sync(body=create_scorer_request, client=self.config.api_client)
 
-        scoreable_node_types = [node_level]
-        version_req = CreateLLMScorerVersionRequest(
```
nachiket-galileo, line 74: Why is this being removed? 
Looks like a breaking change to me

nachiket-galileo (changes requested):

tests/test_metrics.py
jrhone: Because we're setting it on the create scorer request instead to mirror how the UI is sending the requests

src/galileo/metrics.py
jrhone: See comments above. This is a fix.

john-weiler:
Ok yeah this is right. Nice job

PR #338, opened 2025-09-30
feat: Delete metric

jrhone:
**shortcut:** https://app.shortcut.com/galileo/story/40920/python-sdk-no-way-to-delete-a-metric [sc-40920]

**description:** No way to delete a metric in python SDK

### DIFF
--- a/src/galileo/metrics.py
+++ b/src/galileo/metrics.py
@@ -34,9 +34,7 @@ class Metrics:
         self.config = GalileoPythonConfig.get()
 
     def delete_metric(self, name: str) -> None:
-        scorers = Scorers().list()
-        scorers_to_delete = [s for s in scorers if s.name == name]
-
+        scorers_to_delete = Scorers().list(name=name)
         if not scorers_to_delete:
             raise ValueError(f"Scorer with name {name} not found.")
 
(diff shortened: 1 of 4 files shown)


src/galileo/metrics.py
```diff
@@ -28,6 +33,13 @@ class Metrics:
     def __init__(self) -> None:
         self.config = GalileoPythonConfig.get()
 
+    def delete_metric(self, scorer_name: str, scorer_type: ScorerTypes) -> None:
```
john-weiler, line 36: Let's not call this scorer_name ...

Also, let's remove the scorer_type argument. let's delete all metrics that have this name (in practice there will only ever be 1)

john-weiler: let's call it "name"

src/galileo/metrics.py
```diff
@@ -28,6 +33,21 @@ class Metrics:
     def __init__(self) -> None:
         self.config = GalileoPythonConfig.get()
 
+    def delete_metric(self, name: str) -> None:
+        scorers = Scorers().list()
```
john-weiler, line 37: can we just search by name here rather than get all the scorers ?

john-weiler (approved):

PR #350, opened 2025-10-02
feat: Add project user management methods

jrhone:
**shortcut:** https://app.shortcut.com/galileo/story/42295/verizon-2-0-workflow-for-updating-user-permissions-on-projects#activity-42428 [sc-42428]

**description:** Support workflow for updating user permissions on projects

Verizon has workflows that require listing out projects, viewing all users those projects are shared with, and updating those permissions based on their changing org structure and company.


jrhone:
Any chance for a review here?

PR #357, opened 2025-10-06
feat: Set default values for export_records

jrhone:
**shortcut:** https://app.shortcut.com/galileo/story/42678/update-the-export-library

**description:** Set default values for sort, filter and root_type for export_records. The API already sets default for sort and filter but for some reason the generated resources for these aren't optional and have no defaults so we set them to the server defaults. Additionally we set a default for root_type since it's required, and we default to the oldest logstream if no ids are provided.


vamaq (approved):
LGTM!

PR #547, opened 2026-04-14
fix(experiment): preserve user prompt_settings when overriding model

thiagobomfin-galileo:
# User description
**Shortcut:**
[sc-61307](https://app.shortcut.com/galileo/story/61307/fix-silent-data-loss-in-experiment-create-prompt-settings-handling)

**Description:**

`Experiment.create()` had a defensive fallback when both `prompt_settings` and `model` were provided:

```python
settings_dict = (
    effective_prompt_settings.to_dict() if hasattr(effective_prompt_settings, "to_dict") else {}
)
```

If `effective_prompt_settings` ever lacked `to_dict()`, the `else {}` branch silently replaced **all** user-provided settings (temperature, max_tokens, top_p, ...) with an empty dict plus `model_alias`. Since `prompt_settings` is typed as `Optional[PromptRunSettings]` — which always exposes `to_dict()` — the guard was dead defensive code that turned a type contract violation into silent data loss.

This PR removes the `hasattr` guard and calls `.to_dict()` directly. Unexpected types now raise `AttributeError` immediately instead of producing a misconfigured experiment.

Added regression test `test_create_preserves_user_prompt_settings_when_overriding_model_alias` asserting that user-supplied `temperature`, `max_tokens`, and `top_p` survive when `model="GPT-4o"` is also passed.

**Tests:**

- [x] Unit Tests Added

---

# Generated description

Preserve user prompt settings when overriding the model alias by calling <code>PromptRunSettings.to_dict()</code> directly in <code>Experiment.create()</code> so that unexpected inputs raise immediately instead of silently losing fields. Add regression coverage around <code>Experiment.create()</code> to assert that user temperature, max tokens, and top_p survive when model alias is supplied.

<details><summary>Latest Contributors(0)</summary><table><tr><th>User</th><th>Commit</th><th>Date</th></tr></table></details>
This pull request is reviewed by Baz. Review like a pro on <a href=https://baz.co/changes/rungalileo/galileo-python/547?tool=ast>(Baz)</a>.

### DIFF
--- a/src/galileo/experiment.py
+++ b/src/galileo/experiment.py
@@ -434,9 +434,7 @@ class Experiment(StateManagementMixin):
                         "from model parameter (no prompt_settings provided)"
                     )
                 else:
-                    settings_dict = (
-                        effective_prompt_settings.to_dict() if hasattr(effective_prompt_settings, "to_dict") else {}
-                    )
+                    settings_dict = effective_prompt_settings.to_dict()
                     settings_dict["model_alias"] = self.model_alias
                     effective_prompt_settings = PromptRunSettings(**settings_dict)
             elif self._prompt_template is not None and effective_prompt_settings is None:
--- a/tests/test_experiment.py
+++ b/tests/test_experiment.py
@@ -6,7 +6,7 @@ import pytest
 
 from galileo.exceptions import NotFoundError
 from galileo.experiment import Experiment
-from galileo.resources.models import ExperimentResponse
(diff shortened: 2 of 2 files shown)


fernandocorreia-galileo:
Follow-up story created: https://app.shortcut.com/galileo/story/62558/use-promptrunsettings-from-dict-to-preserve-additional-properties-in-experiment-create

fernandocorreia-galileo (approved):

PR #555, opened 2026-04-21
fix(datasets): remap 'output' to 'ground_truth' in get_content() response

thiagobomfin-galileo:
# User description
**Shortcut:**

[sc-62751](https://app.shortcut.com/galileo/story/62751/qa2-0-newsdk-dataset-column-names-are-incorrect)

**Description:**

- `get_content()` was returning `'output'` in `column_names` and row `values_dict` instead of `'ground_truth'`, because the API stores ground_truth as `output` internally
- Added `remap_output_to_ground_truth()` helper in `utils/datasets.py` that acts as the inverse of `normalize_dataset_rows` — applied after every `get_content()` API call

**Tests:**

- [x] Unit Tests Added

---

---

# Generated description

Fix <code>Dataset.get_content</code> and <code>remap_output_to_ground_truth</code> so API-returned <code>output</code> columns are presented as <code>ground_truth</code>, matching dataset expectations. Include focused tests demonstrating column name syncing and row value remapping.

<details><summary>Latest Contributors(2)</summary><table><tr><th>User</th><th>Commit</th><th>Date</th></tr><tr><td>thiago.bomfin@galileo.ai</td><td>fix(dataset): normaliz...</td><td>April 15, 2026</td></tr>
<tr><td>fernando.correia@galil...</td><td>chore: update ruff tar...</td><td>April 07, 2026</td></tr></table></details>
This pull request is reviewed by Baz. Review like a pro on <a href=https://baz.co/changes/rungalileo/galileo-python/555?tool=ast>(Baz)</a>.

### DIFF
--- a/src/galileo/datasets.py
+++ b/src/galileo/datasets.py
@@ -44,7 +44,7 @@ from galileo.resources.models.update_dataset_content_request import UpdateDatase
 from galileo.resources.models.update_dataset_request import UpdateDatasetRequest
 from galileo.resources.types import UNSET, File, Unset
 from galileo.schema.datasets import DatasetRecord
-from galileo.utils.datasets import normalize_dataset_rows, validate_dataset_in_project
+from galileo.utils.datasets import normalize_dataset_rows, remap_output_to_ground_truth, validate_dataset_in_project
 from galileo.utils.exceptions import APIException
 from galileo.utils.log_config import get_logger
 from galileo.utils.projects import resolve_project_id
@@ -92,6 +92,7 @@ class Dataset:
             client=self.config.api_client, dataset_id=self.dataset.id, limit=limit, starting_token=starting_token
         )
 
+        content = remap_output_to_ground_truth(content)
         self.content = content
 
         return content
--- a/src/galileo/datasets.py
+++ b/src/galileo/datasets.py
@@ -94,6 +94,8 @@ class Dataset:
 
         if isinstance(content, DatasetContent):
             content = remap_output_to_ground_truth(content)
+            if not isinstance(content.column_names, Unset):
+                self.dataset.column_names = list(content.column_names)
         self.content = content
 
         return content
--- a/src/galileo/datasets.py
+++ b/src/galileo/datasets.py
@@ -88,11 +88,12 @@ class Dataset:
         if not self.dataset:
             return None
 
-        content: DatasetContent = get_dataset_content_datasets_dataset_id_content_get.sync(
+        content = get_dataset_content_datasets_dataset_id_content_get.sync(
             client=self.config.api_client, dataset_id=self.dataset.id, limit=limit, starting_token=starting_token
         )
 
-        content = remap_output_to_ground_truth(content)
+        if isinstance(content, DatasetContent):
+            content = remap_output_to_ground_truth(content)
         self.content = content
 
         return content
--- a/src/galileo/utils/datasets.py
+++ b/src/galileo/utils/datasets.py
@@ -15,10 +15,12 @@ def remap_output_to_ground_truth(content: DatasetContent) -> DatasetContent:
     The API stores ground_truth as 'output' internally. This is the inverse of
     normalize_dataset_rows, applied when reading content back out.
     """
-    if not isinstance(content.column_names, Unset) and "output" in content.column_names:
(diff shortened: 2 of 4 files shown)


fernandocorreia-galileo (changes requested):
### Issues Found

- **Major: Inconsistency with other dataset-content endpoints.** The fix only covers `Dataset.get_content()`. Sibling methods that also return `DatasetContent`/`DatasetRow` from the same parquet table do **not** apply the remap, so users will still see `output` from them:
  - `Dataset.load_version()` – `src/galileo/datasets.py:162-165` (`GET /datasets/{id}/versions/{index}/content`).
  - `Dataset.extend()` return value – `src/galileo/datasets.py:655-662` re-fetches content and returns rows with `output`. The `__future__` `Dataset.extend()` (`src/galileo/dataset.py:449-497`) forwards those rows to the user.
  - `Dataset.get_version_content()` in `__future__` – `src/galileo/dataset.py:412-447` calls `load_version()`.
  - File: `src/galileo/datasets.py` (lines 162-165, 655-662); `src/galileo/dataset.py` (lines 412-447, 449-497)
  - Suggestion: Apply `remap_output_to_ground_truth` on the results of `load_version()` and the content fetch in `extend()` too. Otherwise the same bug (sc-62751) will reappear under a different entry point and users will see inconsistent schemas between `get_content()` and `get_version_content()`. If out of scope, open a follow-up story.

- **Minor: Mutation semantics differ from the sibling helper.** `normalize_dataset_rows` deliberately copies the caller's dict (`src/galileo/utils/datasets.py:40`) to avoid side effects. `remap_output_to_ground_truth` mutates `content.column_names` by reassignment and mutates `row.values_dict.additional_properties` in place. At the single current call site this is fine because `content` is a freshly-parsed object, but the asymmetric contract is easy to misuse (someone may call this on a cached `self.content` held elsewhere).
  - File: `src/galileo/utils/datasets.py:12-27`
  - Suggestion: Either (a) document "this function mutates the given DatasetContent in place and also returns it" in the docstring, or (b) build a new `DatasetContent` / row copies and return that. Option (a) is cheapest and matches current usage.

- **Nit: Redundant `"output" in content.column_names` guard.** The guard on line 18 short-circuits the list comprehension, but the comprehension itself is already a no-op when `output` isn't present. Dropping the guard simplifies the code without changing behavior. (Not worth blocking on.)
  - File: `src/galileo/utils/datasets.py:18-19`

- **Nit: Stale comment/priority in `convert_dataset_row_to_record`.** The helper documents "Prefer 'output' if both are present (matches API response structure)" and looks up `output` first. On the `get_records_for_dataset` → `dataset.get_content()` path, rows are now remapped, so `output` is never present and the lookup always falls through to `ground_truth`. The function still behaves correctly, but the comment and primary key should be inverted (prefer `ground_truth`, fall back to `output` for raw-API paths like `extend()`/`load_version()` that aren't remapped yet). Also flagged by the Astra review.
  - File: `src/galileo/datasets.py:1084-1098`

- **Nit: `get_content()` crashes if the API returns `HTTPValidationError` (422).** Pre-existing problem — the function is typed `sync() -> Optional[Union[DatasetContent, HTTPValidationError]]` and the old code likewise assigned the 422 object straight into `self.content`. After this PR, `remap_output_to_ground_truth(<HTTPValidationError>)` raises `AttributeError` on `.column_names`. Not introduced here, but this PR makes the surface slightly worse; worth a follow-up to reject non-`DatasetContent` responses.
  - File: `src/galileo/datasets.py:88-98`

### Questions

- Is there an intentional product decision to keep `extend()`/`load_version()` on the "internal" `output` name? If yes, that's worth a comment at the remap call site so future readers know why the coverage is partial. If not, the fix should extend to those callers (see Major item).

## Verdict

**Approve (with a follow-up)**: The premise is verified against `~/projects/api` and the remap is correct for `get_content()`. Strongly recommend expanding the remap to `load_version()` and the content fetch inside `Dataset.extend()` — either in this PR or a same-sprint follow-up — so the schema users see is consistent across dataset read paths.

thiagobomfin-galileo:
@fernandocorreia-galileo Nits adjusted and passed on Astra review

fernandocorreia-galileo (approved):

PR #566, opened 2026-04-23
feat: add agentcontrol logger bridge

namrataghadi-galileo:
# User description
**Shortcut:**
https://app.shortcut.com/galileo/story/62792/galileo-python-add-galileo-logger-bridge-for-agent-control-control-span-ingestion

**Description:**
Adds Agent Control bridge support to GalileoLogger so Agent Control events can be captured as Galileo control spans when the optional dependency is installed. This PR also hardens the integration by:

keeping the import path safe when Agent Control is not installed
supporting the local fallback ControlSpan model in logged trace serialization
preserving the fallback "control" name when callers pass name=None
using a shared trace-context dispatcher so an idle newer logger does not mask an older active logger
Tests:

 Unit Tests Added
 [E2E Test](https://github.com/rungalileo/e2e-testing) Added (if it's a user-facing feature, or fixing a bug) for this PR

---

# Generated description

Enable <code>GalileoLogger</code> to wire into the optional Agent Control SDK by registering a shared <code>GalileoAgentControlBridge</code> that converts Agent Control events into Galileo control spans and manages trace-context dispatching across concurrent loggers. Strengthen control-span modeling by exporting resilient <code>ControlSpan</code>/<code>ControlResult</code> definitions, plumping serialization safeguards, and updating logged-trace handling so fallback spans keep their defaults when the native schema is absent.
<table><tr><th>Topic</th><th>Details</th><tr><td><a href=https://baz.co/changes/rungalileo/galileo-python/566?tool=ast&topic=Control+span+schema>Control span schema</a>
        </td><td>Harden control-span modeling and serialization by exporting resilient <code>ControlAppliesTo</code>/<code>ControlCheckStage</code>/<code>ControlResult</code>/<code>ControlSpan</code> definitions, tightening <code>LoggedControlSpan</code> to string inputs, ensuring schema registries include fallback spans, and guarding serialization against unserializable primitives so fallback spans keep their <code>control</code> defaults while rejecting invalid IDs.<details><summary>Modified files (5)</summary><ul><li>src/galileo/logger/control.py</li>
<li>src/galileo/schema/__init__.py</li>
<li>src/galileo/schema/logged.py</li>
<li>src/galileo/utils/serialization.py</li>
<li>tests/schemas/test_logged.py</li></ul></details><details><summary>Latest Contributors(2)</summary><table><tr><th>User</th><th>Commit</th><th>Date</th></tr><tr><td>namrata.ghadi@galileo.ai</td><td>address comments</td><td>May 04, 2026</td></tr>
<tr><td>fernando.correia@galil...</td><td>chore: update ruff tar...</td><td>April 07, 2026</td></tr></table></details></td></tr>
<tr><td><a href=https://baz.co/changes/rungalileo/galileo-python/566?tool=ast&topic=Tooling+%26+imports>Tooling & imports</a>
        </td><td>Polish development tooling and import hygiene by tightening Ruff hooks/compliance, recording optional global imports without <code>global</code> statements in <code>crewai</code> handlers, and documenting lazy imports/<code>PLC0415</code> allowances for cross-module conditional imports so optional SDKs can be held at arm’s length.<details><summary>Modified files (3)</summary><ul><li>openapi-client-config.yaml</li>
<li>pyproject.toml</li>
<li>src/galileo/handlers/crewai/handler.py</li></ul></details><details><summary>Latest Contributors(2)</summary><table><tr><th>User</th><th>Commit</th><th>Date</th></tr><tr><td>namrata.ghadi@galileo.ai</td><td>Merge branch 'main' in...</td><td>May 11, 2026</td></tr>
<tr><td>ci@rungalileo.io</td><td>chore(release): v2.2.0</td><td>May 06, 2026</td></tr></table></details></td></tr>
<tr><td><a href=https://baz.co/changes/rungalileo/galileo-python/566?tool=ast&topic=Agent+Control+bridge>Agent Control bridge</a>
        </td><td>Bridge Agent Control telemetry into active <code>GalileoLogger</code> sessions by loading the optional SDK, registering the <code>GalileoAgentControlBridge</code> sink/provider, dispatching trace context safely across loggers, exposing the integration via <code>galileo.handlers.agent_control</code>, and covering the lifecycle through new bridge/auto-enable/disable paths plus polished init ordering (including cleanup in <code>terminate</code>). Include the guardrail fixture and regression suites that verify registration ordering, idle logger behavior, event conversion for batch/distributed modes, context matching, and cleanup semantics.<details><summary>Modified files (9)</summary><ul><li>src/galileo/__init__.py</li>
<li>src/galileo/handlers/agent_control/__init__.py</li>
<li>src/galileo/handlers/agent_control/bridge.py</li>
<li>src/galileo/logger/__init__.py</li>
<li>src/galileo/logger/logger.py</li>
<li>tests/conftest.py</li>
<li>tests/test_agent_control_bridge.py</li>
<li>tests/test_logger_batch.py</li>
<li>tests/test_logger_distributed.py</li></ul></details><details><summary>Latest Contributors(2)</summary><table><tr><th>User</th><th>Commit</th><th>Date</th></tr><tr><td>namrata.ghadi@galileo.ai</td><td>Merge branch 'main' in...</td><td>May 11, 2026</td></tr>
<tr><td>ci@rungalileo.io</td><td>chore(release): v2.2.0</td><td>May 06, 2026</td></tr></table></details></td></tr></table>
<sub><a href="https://baz.co/changes/rungalileo/galileo-python/566?tool=ast">Review this PR on Baz</a> | <a href="https://baz.co/agents/baz">Customize your next review</a></sub>


src/galileo/schema/logged.py
```diff
@@ -120,7 +121,8 @@ def _convert_output_to_message(cls, value: LlmSpanAllowedOutputType) -> LoggedMe
 
 # RetrieverSpan and ToolSpan use plain string/document I/O and don't need multimodal widening.
 LoggedSpan = Annotated[
-    LoggedAgentSpan | LoggedWorkflowSpan | LoggedLlmSpan | RetrieverSpan | ToolSpan, Field(discriminator="type")
+    LoggedAgentSpan | LoggedWorkflowSpan | LoggedLlmSpan | RetrieverSpan | ToolSpan | ControlSpan,
+    Field(discriminator="type"),
```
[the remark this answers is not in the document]
namrataghadi-galileo: this is now fixed

src/galileo/handlers/agent_control/bridge.py
```diff
@@ -0,0 +1,302 @@
...
+
+        return _ActiveContext(trace_id=str(root_parent.id), span_id=str(current_parent.id))
+
+    @staticmethod
+    def _matches_active_context(event: Any, active_context: _ActiveContext) -> bool:
+        event_trace_id = _normalize_context_id(getattr(event, "trace_id", None))
+        event_span_id = _normalize_context_id(getattr(event, "span_id", None))
+        active_trace_id = _normalize_context_id(active_context.trace_id)
+        active_span_id = _normalize_context_id(active_context.span_id)
+        return event_trace_id == active_trace_id and event_span_id == active_span_id
```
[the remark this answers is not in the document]
fercor-cisco: - **Analysis**: Valid, but very small. Each `uuid.UUID(text)` parse is sub-microsecond, so the redundancy is ~1µs per event. The bigger argument for fixing it is *clarity*: it suggests both sides need normalization, which isn't true. The simplification is a one-liner.
  - **Recommended action**: Drop the two redundant calls — only normalize the event-side IDs and compare to the already-canonical `active_context.trace_id`/`span_id`. Don't add `# type: ignore` or other indirection.

src/galileo/logger/logger.py
```diff
@@ -1791,6 +1835,86 @@ def add_agent_span(
...
+        }
+        if name is not None:
+            span_kwargs["name"] = name
+
+        span = ControlSpan(**span_kwargs)
+        span._parent = current_parent
+        self.add_child_span_to_parent(span)
+
+        if self.mode == "distributed":
+            self._ingest_step_streaming(span)
```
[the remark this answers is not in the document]
fercor-cisco: - **Analysis**: The mismatch is real but `_ingest_step_streaming` is already called via `@warn_catch_exception` and only reads `.type`/`.id`/runs `copy.deepcopy(span)`, so runtime is fine. Native `ControlSpan` from galileo-core (when present) DOES inherit from `Span`, so this only affects fallback builds. Importantly, `mypy --strict` is not enforced on this file today (logger.py already has 19 errors on origin/main without this change, and this PR introduces no new ones at lines 1898-1903 — verified locally).
  - **Recommended action**: Don't widen `_ingest_step_streaming`'s type signature for a fallback path. Keep things simple: leave a one-line comment at `logger.py:1902` noting the fallback `ControlSpan` is structurally compatible with `Span` for streaming, and rely on the native `ControlSpan` inheriting from `Span` once core ships it. No `# type: ignore` needed because mypy doesn't error here today.

src/galileo/handlers/agent_control/bridge.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+def _normalize_context_id(value: Any) -> str | None:
+    """Return a canonical UUID string, or None if value is absent or not a valid UUID."""
+    if value is None:
+        return None
+
+    text = str(value).strip()
+    if not text:
+        return None
```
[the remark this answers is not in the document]
fercor-cisco: - **Analysis**: Real but currently theoretical. The state is only mutated when `agent_control` is importable, and the bridge tests are the only ones that fake those modules in. Other tests will silently take the `ImportError` path in `_auto_enable_agent_control_if_available`. Still, an autouse cleanup fixture in `tests/conftest.py` would close the gap cheaply.
  - **Recommended action**: Defer to a follow-up. If you choose to fix it now, add a `tests/conftest.py` autouse fixture that, in teardown, calls `_REGISTERED_BRIDGES.clear()` and resets `_PREVIOUS_TRACE_CONTEXT_PROVIDER` only if the bridge module is loaded. Don't move to a formal singleton — that's over-engineering for state that's only meaningful when an optional package is installed.

pyproject.toml
fercor-cisco, line 1: `agent-control` is not declared as an optional dependency. `pyproject.toml:332` declares extras for `langchain`, `openai`, `crewai`, `middleware`, `otel`, `all` — but not `agent-control`. Users discovering the bridge via `from galileo import setup_agent_control_bridge` have no obvious install path; they'd have to know to `pip install agent-control` (and now also `agent-control-telemetry`). The bridge silently doesn't auto-enable if either is missing, so problems surface late.
  - File: `pyproject.toml` (lines 332-338)
  - Suggestion: Add `agent-control = ["agent-control (>=X.Y.Z)", "agent-control-telemetry (>=A.B.C)"]` and include it in `all`. (Use whatever lower bounds match the APIs you depend on — the bridge needs `register_control_event_sink`, `unregister_control_event_sink`, `agent_control_telemetry.sinks.SinkResult`, and `agent_control_telemetry.trace_context.set_trace_context_provider`. I confirmed these exist on `agentcontrol/agent-control` `origin/main` at `sdks/python/src/agent_control/observability.py:853` and `telemetry/src/agent_control_telemetry/trace_context.py:18`.)

src/galileo/logger/logger.py
```diff
@@ -2169,6 +2282,8 @@ def terminate(self) -> None:
                     # Event loop might be closed during shutdown, log warning but don't crash
                     self._logger.warning(f"Could not flush during terminate due to event loop shutdown: {e}")
         finally:
+            self.disable_agent_control()
```
fercor-cisco: `disable_agent_control()` runs in `GalileoLogger.terminate`'s `finally` (`logger.py:2285`) but is not wrapped in `try/except`. During interpreter shutdown atexit fires `terminate()`, and at that point `agent_control_telemetry.trace_context` may already be partially torn down. `bridge.unregister()` then calls `self._modules.agent_control.unregister_control_event_sink(self._sink)` and `self._modules.trace_context.set_trace_context_provider(...)`. If those raise, the `finally` aborts before the task-handler shutdown below it runs, which is exactly what that finally was added to guarantee.
  - File: `src/galileo/logger/logger.py` (line 2285)
  - Suggestion:
    ```python
    finally:
        try:
            self.disable_agent_control()
        except Exception as exc:
            self._logger.warning("GalileoLogger.terminate: agent control unregister failed: %s", exc)
        # ... existing task-handler cleanup
    ```

src/galileo/handlers/agent_control/bridge.py
```diff
@@ -0,0 +1,302 @@
...
+            return None
+
+        return _ActiveContext(trace_id=str(root_parent.id), span_id=str(current_parent.id))
+
+    @staticmethod
+    def _matches_active_context(event: Any, active_context: _ActiveContext) -> bool:
+        event_trace_id = _normalize_context_id(getattr(event, "trace_id", None))
+        event_span_id = _normalize_context_id(getattr(event, "span_id", None))
+        active_trace_id = _normalize_context_id(active_context.trace_id)
+        active_span_id = _normalize_context_id(active_context.span_id)
```
fercor-cisco: Redundant `_normalize_context_id` calls in `_matches_active_context` (`bridge.py:253-258`). `_active_context()` already builds canonical UUID strings via `str(root_parent.id)` / `str(current_parent.id)`. Re-normalizing those is a no-op and slightly misleads readers about whether both sides need normalization. (Same as the open astra-bot thread above.)
  - File: `src/galileo/handlers/agent_control/bridge.py` (lines 252-258)
  - Suggestion: Normalize only the event-side IDs:
    ```python
    @staticmethod
    def _matches_active_context(event: Any, active_context: _ActiveContext) -> bool:
        event_trace_id = _normalize_context_id(getattr(event, "trace_id", None))
        event_span_id = _normalize_context_id(getattr(event, "span_id", None))
        return event_trace_id == active_context.trace_id and event_span_id == active_context.span_id
    ```

src/galileo/handlers/agent_control/bridge.py
```diff
@@ -0,0 +1,302 @@
...
+def _extract_control_input(metadata: dict[str, Any] | None) -> str:
+    if not metadata:
+        return ""
+
+    for key in ("selected_data", "selected_value", "value", "input"):
+        if key in metadata and metadata[key] is not None:
+            value = metadata[key]
+            return value if isinstance(value, str) else serialize_to_str(value)
+
+    return ""
```
fercor-cisco, line 99: minor: `_extract_control_input` silently coerces missing-input to `""` (`bridge.py:90-99`). The downstream `ControlSpan.input` accepts any value, so an empty string vs `None` is a UX nuance — empty strings will appear as blank inputs in the UI even when the upstream Agent Control event genuinely had no representative input. Consider `None` to make "no input" distinguishable from "empty input".
  - File: `src/galileo/handlers/agent_control/bridge.py` (lines 90-99)
  - Suggestion: Return `None` (and let the caller / model defaults handle it). Low priority.

src/galileo/logger/logger.py
```diff
@@ -256,6 +261,7 @@ def __init__(
             self.log_stream_name = log_stream
             if local_metrics:
                 self.local_metrics = local_metrics
+            self._auto_enable_agent_control_if_available()
             atexit.register(self.terminate)
             return
```
fercor-cisco, line 266: **Nit**: The ingestion-hook constructor path (`logger.py:259-266`) calls `_auto_enable_agent_control_if_available()` *before* `atexit.register(self.terminate)`. If auto-enable raises and ImportError handling fails to swallow it, the atexit hook never gets registered and the bridge remains active. The current code catches `ImportError` and `Exception`, so this is theoretical, but the standard logger path also calls auto-enable AFTER `_init_distributed_trace_stubs` and BEFORE atexit. Minor inconsistency.
  - File: `src/galileo/logger/logger.py` (lines 264-265, 353-356)
  - Suggestion: Move `_auto_enable_agent_control_if_available()` to AFTER `atexit.register(self.terminate)` in both branches. Then even a partial init still teardown-cleans the bridge on interpreter exit.

pyproject.toml
namrataghadi-galileo: This is done on purpose. We have explicitly decided to **NOT** add a dependency on agent-control in galileo-sdk. The expectation is that the Agent Developers will explicitly install agent-control in their environment. Since agent control is optional, . This is intentional. We don’t want galileo to declare or pull in agent-control as an SDK extra because Agent Control is meant to be installed independently in environments that actually use it. The Galileo bridge is opportunistic: it checks whether both agent_control and agent_control_telemetry are importable, and only then registers the sink/provider needed for control span ingestion.

src/galileo/handlers/agent_control/bridge.py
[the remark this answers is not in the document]
namrataghadi-galileo: resolved

src/galileo/handlers/agent_control/bridge.py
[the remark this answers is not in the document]
namrataghadi-galileo: resolved

src/galileo/logger/logger.py
namrataghadi-galileo: addressed

src/galileo/handlers/agent_control/bridge.py
namrataghadi-galileo: addressed

src/galileo/handlers/agent_control/bridge.py
namrataghadi-galileo: addressed

src/galileo/logger/logger.py
namrataghadi-galileo: addressed

src/galileo/logger/logger.py
```diff
@@ -1791,6 +1823,86 @@ def add_agent_span(
...
+            "agent_name": agent_name,
+            "check_stage": check_stage,
+            "applies_to": applies_to,
+            "evaluator_name": evaluator_name,
+            "selector_path": selector_path,
+        }
+        if name is not None:
+            span_kwargs["name"] = name
+
+        span = ControlSpan(**span_kwargs)
```
pratduv: Wondering if we need a `LoggedControlSpan` type here, to match the other types `LoggedAgent/Trace/Workflow/ToolSpan` types.

The idea here is that we have different read and write type schemas (which is why we have the LoggedXYZTypes). Types that aren't multimodal also have been cloned (like the LoggedToolSpan is the same as ToolSpan (on read), but we still have a LoggedToolSpan type).

Also I just looked at the ingest types for ControlSpan types in both the ingest service and the SDK, and I think the input types may be overqualified.

The input type is supposed to be just "str" right? So in BaseControlSpan we could override input: str like Tool/Retriever do, then ripple downstream.

pratduv (changes requested):
Excited to have agent control in the logger soon!

couple comments - 

1. What's parent_id on a control span supposed to be? The BaseControlSpan docstring in core says it's the LLM/tool span id. But it looks like that can't actually happen here because LLM and tool spans are leaves in our logger, they don't push themselves onto the parent stack via _attach_parentable_span (only workflow and agent spans do). So current_parent() returns the wrapping workflow/agent during an LLM call, never the LLM span itself, and that ends up as the control span's parent_id. The test assert control_span.parent_id == workflow.id confirms it's the workflow, not the LLM span.

So either the docstring is wrong and control spans really are children of the wrapping workflow/agent (and we should update the core docstring), or the integration needs to do something different so callers can pass the LLM/tool span id explicitly. 



2.The four id-ish fields on the local ControlSpan in this PR (id, session_id, trace_id, parent_id) are typed as Any | None. The matching ControlSpan in galileo-core uses Optional[UUID4]. So this local copy accepts random objects (ints, dicts, strings) that the real one rejects. Tests run against the local copy (mocked imports force the fallback path), production runs against the galileo-core version, so anything weird that slips through here only blows up in prod. Tighten these to UUID | str | None so the boundary catches it. The string variant is fine since add_control_span already parses string ids to UUID.

3. re: the unregister flow restoring _PREVIOUS_TRACE_CONTEXT_PROVIDER when the last bridge exits., we snapshot whatever provider was already installed at the moment the first Galileo bridge registers, and stash it in _PREVIOUS_TRACE_CONTEXT_PROVIDER. While our bridges are active, nothing stops some other code from calling agent_control_telemetry.trace_context.set_trace_context_provider(provider_B). When the last Galileo bridge unregisters, we put back the snapshot and silently overwrite provider_B. So whoever installed provider_B loses their tracing hook without any warning. This is less of a concern but i'm not sure how the threading/ context var clobbering works here.

pratduv (commented):
There may be a contract issue in `logger.add_llm_span(...)`

_active_context() in the bridge reads only _target_context.get(), and there's no parent-stack fallback.

logger.agent_control_span_context() exists and IS the manual public API (it's already what the auto-instrumented paths call internally).

But add_llm_span's docstring (and add_tool_span/add_protect_tool_span) doesn't mention this anywhere. The new id= parameter is documented as "Explicit span ID to use instead of generating a new one"

So a user calling logger.add_llm_span(...) directly will get no Agent Control attribution and have no way to know that's happening.

pratduv (approved):
Read through the thread, agree with the sibling-under-workflow structure given the execution flow argument.  Approved!

Few comments on tests/1 nit

Two of the type-boundary tests got nuked along with the round-2/3 changes - `test_fallback_control_span_rejects_non_uuidish_id_fields` and `test_logged_control_span_rejects_non_string_input`. The schemas they covered (UUID | str | None on the fallback id-ish fields, str on LoggedControlSpan.input) are still strict, just untested atm. 

The reload-and-restore in test_logged_trace_roundtrip_with_fallback_control_span and test_add_control_span_uses_model_default_name_in_fallback_mode is still outside the with monkeypatch.context() block. If any inner assert fails, the unwind reloads never run and the rest of the xdist worker sees corrupted module state. Maybe move them into a try/finally?

add_control_span is annotated -> LoggedControlSpan but @nop_sync and @warn_catch_exception(exceptions=(Exception,)) can both return None (logging disabled, or any exception caught — including the no-active-trace case where add_child_span_to_parent raises). We could tighten the annotation to LoggedControlSpan | None

PR #568, opened 2026-04-28
fix: the filter schema patches

namrataghadi-galileo:
**Shortcut:**
https://app.shortcut.com/galileo/story/63570/import-openapi-yaml-sh-filter-schema-patches-are-stale-andnode-ornode-notnode-filterexpression-filterleaf

**Description:**

Update scripts/import-openapi-yaml.sh so the five hardcoded filter-schema patch keys match the current OpenAPI schema names after LogRecordsFullyAnnotatedFilter was added to the discriminated union. This restores the title-shortening patch used to avoid Windows filename-length issues during API client generation, and prevents yq from creating orphan schema entries in openapi.yaml.

**Tests:**

 Unit Tests Added
 [E2E Test](https://github.com/rungalileo/e2e-testing) Added (if it's a user-facing feature, or fixing a bug)

**Validation performed:**

Ran ./scripts/import-openapi-yaml.sh http://127.0.0.1:8088/client against a local API
Confirmed the five target schemas in openapi.yaml have shortened titles
Confirmed stale orphan schema keys were not created
Ran ./scripts/auto-generate-api-client.sh and did not hit the Windows-style filename length failure


jonahe (approved):
Nice

openapi.yaml
```diff
@@ -3,8 +3,8 @@ info:
   title: FastAPI
   version: 0.1.0
 servers:
-  - url: https://api.galileo.ai
-    description: Galileo Client APIs - galileo-v2
+  - url: http://localhost:8088
+    description: Galileo Client APIs - local
```
fercor-cisco, line 7: Created https://app.shortcut.com/galileo/story/63841/add-lint-rule-for-openapi-yaml-server-url to catch this in the future

fernandocorreia-galileo:
> ⚠️ **Astra encountered an error while reviewing this PR.**

An unexpected error occurred while running the review. Job: [`unknown`](https://console.cloud.google.com/logs/query;query=resource.type%20%3D%20%22cloud_run_job%22%20resource.labels.job_name%20%3D%20%22astra-job%22%20resource.labels.location%20%3D%20%22us-west1%22%20labels.%22run.googleapis.com/execution_name%22%20%3D%20%22unknown%22%20severity%3E%3DDEFAULT;storageScope=project?project=rungalileo-dev)

openapi.yaml
fercor-cisco, line 7: **🟠 major** (bug): The server URL and description were changed from the production endpoint to `localhost:8088` because the spec was regenerated from a local dev server. This should remain pointing to the production API (`https://api.galileo.ai`). The `import-openapi-yaml.sh` script should normalize this as part of its post-processing patches.

```suggestion
  - url: https://api.galileo.ai
    description: Galileo Client APIs - galileo-v2
```

fercor-cisco (commented):
> ⚠️ **This review was generated by an AI agent (Astra) and may contain mistakes. Please verify all suggestions independently.**

**Verdict: request_changes** — The openapi.yaml server URL was changed from the production URL (https://api.galileo.ai) to http://localhost:8088 because the script was run against a local API server, and the result was committed as-is.

## General Comments
- **🟠 major** (bug): The `openapi.yaml` was regenerated from a local dev server (`http://127.0.0.1:8088/client` per the PR description), and the resulting `servers[0].url` was committed as `http://localhost:8088` instead of `https://api.galileo.ai`. While the auto-generated Python client code doesn't embed this URL (the SDK uses `GALILEO_CONSOLE_URL` at runtime), the committed spec is now incorrect metadata. This should be fixed by either: (1) adding a post-processing step to `import-openapi-yaml.sh` that normalizes the server URL back to production, or (2) regenerating from the production API. The reviewer already created sc-63841 for a lint rule, but the committed spec should still be corrected.
- **🟡 minor** (design): The script fix itself (adding `LogRecordsFullyAnnotatedFilter__` to the five hardcoded schema key names) is correct and matches the current API schema. However, as noted in the Shortcut story, these hardcoded keys are inherently fragile — any future addition to the filter discriminated union will silently break the patches in the same way (yq creates orphan entries instead of failing). Consider making `import-openapi-yaml.sh` resilient by matching schema keys via a pattern (e.g., `keys | select(startswith("AndNode_Annotated_Union_"))`) rather than exact match, or at minimum adding a validation step that errors if any of the five target schemas still have their original long titles after patching.

## Follow-ups
Suggested follow-up work that could be tracked as Shortcut stories:
- `scripts/import-openapi-yaml.sh:49-49`: The five hardcoded filter schema keys in the yq patch command are fragile — any future change to the filter discriminated union will silently break them (yq creates missing keys instead of failing). Consider using pattern-based matching (e.g., `keys | select(startswith(...))`) or adding a post-patch validation step that checks the target schemas actually have shortened titles. This is tracked in the Shortcut story as a long-term improvement.

openapi.yaml
fercor-cisco: Good catch and thanks for creating the tracking ticket. Beyond a lint rule, the `import-openapi-yaml.sh` script itself should normalize the server URL as part of its existing yq patch pipeline (e.g., `.servers[0].url = "https://api.galileo.ai" | .servers[0].description = "Galileo Client APIs - galileo-v2"`). That way the committed spec is always correct regardless of which server the script was run against.

fercor-cisco, line 7: **🟠 major** (bug): The server URL and description were changed from the production endpoint (`https://api.galileo.ai`, `Galileo Client APIs - galileo-v2`) to localhost values because the spec was regenerated from a local dev server. This should be reverted to the production values. Additionally, the `import-openapi-yaml.sh` script should add a yq patch to normalize the server URL so this can't happen again regardless of which server the script is pointed at.

```suggestion
  - url: https://api.galileo.ai
    description: Galileo Client APIs - galileo-v2
```

scripts/import-openapi-yaml.sh
```diff
@@ -37,16 +37,16 @@ curl -s "${HOST_URL}/openapi.json" | poetry run python -c 'import sys, json, yam
...
...sCollectionFilter__LogRecordsTextFilter__LogRecordsFullyAnnotatedFilter___FieldInfo_annotation_NoneType__required_True__discriminator__type____"].title = "FilterExpressionLogRecordsFilter" | .components.schemas["FilterLeaf_Annotated_Union_LogRecordsIDFilter__LogRecordsDateFilter__LogRecordsNumberFilter__LogRecordsBooleanFilter__LogRecordsCollectionFilter__LogRecordsTextFilter__LogRecordsFullyAnnotatedFilter___FieldInfo_annotation_NoneType__required_True__discriminator__type____"].title = "FilterLeafLogRecordsFilter" | .components.schemas["NotNode_Annotated_Union_LogRecordsIDFilter__LogRecordsDateFilter__LogRecordsNumberFilter__LogRecordsBooleanFilter__LogRecordsCollectionFilter__LogRecordsTextFilter__LogRecordsFullyAnnotatedFilter___FieldInfo_annotation_NoneType__required_True__discriminator__type____"].title = "NotNodeLogRecordsFilter" | .components.schemas["OrNode_Annotated_Union_LogRecordsIDFilter__LogRecordsDateFilter__LogRecordsNumberFilter__LogRecordsBooleanFilter__LogRecordsCollectionFilter__LogRecordsTextFilter__LogRecordsFullyAnnotatedFilter___FieldInfo_annotation_NoneType__required_True__discriminator__type____"].title = "OrNodeLogRecordsFilter"'  "$HOME_DIR/openapi.yaml"
```
fercor-cisco, line 49: **🟠 major** (bug): The yq patch pipeline should also normalize the server URL to prevent committing a localhost URL when the script is run against a local dev server. Add a patch at the beginning of the pipeline:
```
.servers[0].url = "https://api.galileo.ai" | .servers[0].description = "Galileo Client APIs - galileo-v2" | ...
```
This is the root cause of the server URL regression - without this normalization, anyone who runs the script against a non-production server will commit the wrong URL.

fercor-cisco (commented):
> ⚠️ **This review was generated by an AI agent (Astra) and may contain mistakes. Please verify all suggestions independently.**

**Verdict: request_changes** — The server URL in openapi.yaml was changed from the production endpoint to localhost:8088, which was merged into main. This needs a follow-up fix.

## General Comments
- **🟠 major** (bug): The `openapi.yaml` committed in this PR has the server URL changed from `https://api.galileo.ai` (production) to `http://localhost:8088` (local dev). This happened because the script was run against a local server and the resulting spec was committed as-is. While the generated client doesn't hardcode the base URL (it's passed at construction time), the spec file itself is incorrect documentation and could confuse downstream consumers or other code generators.

The `import-openapi-yaml.sh` script should be updated to normalize the server URL as part of its yq patch pipeline, e.g.:
```
.servers[0].url = "https://api.galileo.ai" | .servers[0].description = "Galileo Client APIs - galileo-v2"
```
This ensures the committed spec is always correct regardless of which server the script was run against. I see this was already noted in a review comment and a tracking ticket (sc-63841) was created, but the PR was merged with the incorrect URL.

## Follow-ups
Suggested follow-up work that could be tracked as Shortcut stories:
- `scripts/import-openapi-yaml.sh:49-49`: The filter schema patch keys are brittle - any future addition to the discriminated union will silently break them. Consider making the patch resilient by matching titles via a pattern/regex rather than hardcoded fully-qualified schema keys (as noted in the Shortcut ticket sc-63570).

openapi.yaml
fercor-cisco: Agree with this finding. The server URL should remain `https://api.galileo.ai` in the committed spec. The suggested fix to revert the values is correct, but the script itself should also be updated to normalize the server URL as part of its yq patch pipeline (as noted in the follow-up comment) to prevent this class of issue from recurring.

fercor-cisco: Strongly agree. Adding server URL normalization to the yq patch pipeline is the proper fix. A lint rule (sc-63841) is a good safety net, but normalizing in the script is the primary defense since it prevents the problem at the source rather than catching it after the fact.

PR #573, opened 2026-04-29
fix: schema patches

thiagobomfin-galileo:
**Shortcut:**

**Description:**

- Revert #568
- Re add the fix for the `scripts/import-openapi-yaml.sh`
- Generated api client using the `scripts/import-openapi-yaml.sh` and `scripts/auto-generate-api-client.sh` scripts
- Fix issues in some Jinja templates that were blocking the `scripts/import-openapi-yaml.sh` and `scripts/auto-generate-api-client.sh` scripts from running.

The changes in the files were done by the ruff dependency, and the difference that needed to be reverted was happening because of a mismatch of the local installed ruff version and the version that was defined on pyproject.toml, this was solved by running locally poetry install to reinstall the dependencies and after that running `./scripts/auto-generate-api-client.sh`

**Tests:**

- [ ] Unit Tests Added


namrataghadi-galileo (approved):

Murike (approved):
lgtm

fercor-cisco (approved):

